PR #3, opened 2026-04-16
fix(build): copy torch DLLs and relax rank_bm25 constraint for CI

zaxbysauce:
## Summary

PyInstaller one-dir builds on Windows need torch DLLs (c10.dll, torch.dll, etc.) in the same directory as the executable, not just in _internal/. This fixes runtime model loading failures in the distributed .exe.

## Dependency Fix

This PR also relaxes the 
ank_bm25 version constraint in requirements.txt from >=0.2.3 to >=0.2.2. Version 0.2.3 does not exist on PyPI, causing CI pip install to fail. This blocks installation of fastapi, customtkinter, and other dependencies needed for test execution.

## Changes

1. **scripts/build.py**: Copy torch DLLs to root app directory for Windows one-dir builds
2. **requirements.txt**: Relax rank_bm25 constraint from >=0.2.3 to >=0.2.2

## Test Status

The dependency fix enables CI to install all required packages. Tests that previously failed due to missing modules (fastapi, customtkinter) should now be able to collect and run.


zaxbysauce:
Reviewed

Issue #25, opened 2026-07-02
[web_ui PR-6/7] UI/Visual/Theming &amp; Accessibility Remediation (fixes "the UI is messed up")


zaxbysauce:
## Context

This is PR **6 of 7** from a full audit of `web_ui/` (see `WEB_UI_AUDIT_REPORT.md` on `claude/html-app-audit-rag-1uszpr`). This PR closes the user complaint **"the UI is messed up"** plus the full WCAG 2.1 AA accessibility sweep. Note: issue #18 (a separate, earlier UI/UX audit) covers an overlapping-but-distinct set of look-and-feel gaps (typography scale, transcript width, composer styling, sidebar redesign) — read it before starting, since fixing both audits' findings in the same pass on shared files (tokens.css, ChatMessageBubble.tsx, ChatInput.tsx, NavigationRail.tsx) will be more efficient than two separate passes. Feel free to close #18 as part of this PR if you address its findings too, or leave it open and reference this PR if you don't get to all of it.

**Prerequisite:** PR-1 (#20) must merge first.

## Scope — explicit exclusions to avoid conflicts

Files this PR owns: `web_ui/src/styles/theme.css`, `web_ui/src/styles/tokens.css`, `web_ui/src/layouts/AppLayout.tsx`, `web_ui/src/App.tsx`, `web_ui/src/components/ChatMessageBubble.tsx`, `web_ui/src/components/ChatMessageList.tsx`, `web_ui/src/components/ChatInput.tsx`, `web_ui/src/components/DropZone.tsx` (visual/CSS-token concerns only — file-validation logic is PR-4's), `web_ui/src/components/NavigationRail.tsx`, `web_ui/src/components/SourceCitation.tsx` (popover styling only — citation resolution logic is PR-3's), `web_ui/src/components/ToastProvider.tsx`, `web_ui/src/components/Toast.tsx`, `web_ui/src/components/EmptyState.tsx`, `web_ui/src/components/LoadingSkeleton.tsx`, `web_ui/src/components/MarkdownRenderer.tsx`, `web_ui/src/components/ErrorBoundary.tsx` (mounting it — the component itself is already implemented), `web_ui/src/components/ModelDownloadProgress.tsx`, `web_ui/src/components/StreamingIndicator.tsx`, `web_ui/src/components/DocumentList.tsx` (visual/a11y only — data/deletion logic is PR-4's), `web_ui/index.html`.

[...]

Issue #23, opened 2026-07-02
[web_ui PR-4/7] Document Ingestion &amp; Storage Integrity (fixes disappearing/corrupted knowledge base)


zaxbysauce:
## Context

This is PR **4 of 7** from a full audit of `web_ui/` (see `WEB_UI_AUDIT_REPORT.md` on `claude/html-app-audit-rag-1uszpr`). This PR closes the defects that make the document corpus unreliable: it disappears on browser restart, silently fails to index, and can be corrupted by ordinary user actions (delete-while-processing, duplicate upload, rapid navigation).

**Prerequisite:** PR-1 (#20) must merge first.

## Scope

Files this PR owns: `web_ui/src/pages/DocumentsPage.tsx`, `web_ui/src/components/DropZone.tsx` (file-handling/validation logic — the drag-highlight CSS token fix belongs to PR-6), `web_ui/src/components/DocumentList.tsx` (data/status logic — visual/contrast fixes belong to PR-6), `web_ui/src/lib/processing/*` (all extractors, `text-chunker.ts`), `web_ui/src/lib/storage/document-store.ts`, `web_ui/src/lib/search/vector-index.ts` (persistence/storage-namespace concerns — the text-loss-on-search fix belongs to PR-3), `web_ui/src/lib/search/keyword-index.ts`, `web_ui/src/lib/embeddings/memory-aware.ts`, `web_ui/vite.config.ts` (the edgevec snippet stub only — build/base-path settings belong to PR-1).

**Shared-file note:** PR-3 (#22) needs a small addition to this PR's territory — capturing filename/page number on chunks at indexing time (`processFile` in `DocumentsPage.tsx`) so real citations are possible. Land that specific field-capture as part of your work here (it's a natural fit alongside the storage-namespace fix below, since you're already touching how chunks get their metadata) and flag it clearly in your PR description so PR-3 can consume it.

### 1. Per-session random storage namespace destroys persistence and isolates tabs (CRITICAL)
[...]

Issue #22, opened 2026-07-02
[web_ui PR-3/7] RAG Retrieval Grounding &amp; Citation Accuracy (fixes hallucinated / ungrounded answers)


zaxbysauce:
## Context

This is PR **3 of 7** from a full audit of `web_ui/` (see `WEB_UI_AUDIT_REPORT.md` on `claude/html-app-audit-rag-1uszpr` for the full report and verification appendix). This PR closes the retrieval-quality defects that make answers unreliable and unverifiable — the single most important cluster for a tier-0 EHR support tool, since ungrounded/uncited answers are a patient-safety-adjacent risk.

**Prerequisite:** PR-1 (#20) must merge first.

## Scope

Files this PR owns: `web_ui/src/lib/rag/rag-orchestrator.ts` (retrieval pipeline: `query()`, `buildContext`, `buildMessages` — see shared-ownership note below), `web_ui/src/lib/rag/rag-presets.ts`, `web_ui/src/lib/embeddings/embedding-service.ts`, `web_ui/src/lib/search/reranker.ts`, `web_ui/src/lib/search/rrf-fusion.ts`, `web_ui/src/lib/search/vector-index.ts` (the text-loss fix specifically — vector-index's storage/persistence concerns otherwise belong to PR-4), `web_ui/src/components/SourceCitation.tsx` (citation *logic*/resolution — the popover *styling* overflow fix belongs to PR-6).

**Shared-file notes:**
- `rag-orchestrator.ts`'s constructor (line ~107, `opts?.llmService ?? WebLLMService.getInstance()`) is also touched by PR-2 (#21) for the default-engine fix — coordinate or just change the one line yourself and mention it in your PR description so PR-2 doesn't duplicate it.
- The metadata capture needed for real citations (finding #7 below) requires a small change to `web_ui/src/pages/DocumentsPage.tsx`'s chunk-creation code, which PR-4 (#23) owns. **This PR depends on that change landing** — either wait for PR-4, or coordinate to land the minimal metadata field first. Don't duplicate PR-4's broader ingestion work.

### 1. Vector-retrieved passages lose their text — the LLM is grounded on placeholder strings (CRITICAL)
[...]

Issue #21, opened 2026-07-02
[web_ui PR-2/7] Chat Engine Initialization &amp; Session Lifecycle (fixes "chat sometimes doesn't work")


zaxbysauce:
## Context

This is PR **2 of 7** from a full audit of `web_ui/` (see `WEB_UI_AUDIT_REPORT.md` on `claude/html-app-audit-rag-1uszpr` for the full report, including a verification appendix — every finding here was independently re-traced by a fresh adversarial reviewer and confirmed real). This PR closes the root causes of the user complaint **"chat sometimes doesn't work."**

**Prerequisite:** PR-1 (build/weights/packaging, #20) must merge first — you cannot verify any of this against a repo that doesn't compile or has no model weights.

## Scope

Files this PR owns: `web_ui/src/pages/ChatPage.tsx` (engine lifecycle / init / readiness-overlay concerns only — see note below on shared ownership), `web_ui/src/lib/llm/llm-factory.ts`, `web_ui/src/lib/llm/wllama-service.ts`, `web_ui/src/lib/llm/web-llm-service.ts`, `web_ui/src/lib/llm/webgpu-watchdog.ts`, `web_ui/src/lib/llm/readiness-gate.ts`, `web_ui/src/lib/llm/model-readiness.ts`, `web_ui/src/lib/inference/InferenceModeContext.tsx`, `web_ui/src/hooks/useKeyboardShortcuts.ts`, `web_ui/src/hooks/useServiceInitialization.ts`, `web_ui/src/lib/api/streaming.ts`, `web_ui/src/lib/api/auth.ts`, `web_ui/src/lib/api/client.ts`.

**Shared-file note:** `web_ui/src/pages/ChatPage.tsx` is also touched by PR-6 (#25) for autoscroll, focus management, dialog semantics, aria-live — purely visual/a11y concerns — and by PR-3 (#22) for a 1-line default in `rag-orchestrator.ts`'s constructor. Recommend this PR (#21) lands first since it's higher severity; PR-6 will rebase onto it.

### 1. Browser LLM is never initialized — browser-local chat fails on every send (CRITICAL)
[...]

PR #27, opened 2026-07-12
fix(web_ui): build, weights, and offline-packaging integrity (Issue #20)

### LINKED ISSUE #20: [web_ui PR-1/7] Build, Weights &amp; Offline-Packaging Integrity (BLOCKER — merge first)
## Context

This is PR **1 of 7** from a full audit of `web_ui/` (see `WEB_UI_AUDIT_REPORT.md` on branch `claude/html-app-audit-rag-1uszpr` for the complete report with full verification appendix — 174 raw findings → 140 deduped → 86 adversarially verified: 68 confirmed, 16 adjusted, 0 refuted). This PR is a **hard blocker**: none of the other 6 PRs (chat engine, RAG retrieval, document ingestion, settings, UI/a11y, safety epic) can be meaningfully tested until the repo compiles, has real model weights, and its own readiness checks stop lying.

**Do not start PRs 2–6 work against this repo until this PR is merged** — you'll be debugging symptoms of these root causes instead of your own feature.

## Scope

Files this PR owns (exclusively): `web_ui/src/lib/processing/pdf-extractor.ts`, `web_ui/src/lib/processing/xlsx-extractor.ts`, `web_ui/src/lib/models/model-manifest.ts`, `web_ui/scripts/*`, `web_ui/public/models/manifest.json`, `web_ui/vitest.config.ts`, `web_ui/vite.config.ts` (base-path/sourcemap settings only — the edgevec snippet stub itself is owned by PR-4), `web_ui/PACKAGING.md`, root `models/` weight staging.

### 1. The checked-in `web_ui/` does not compile (CRITICAL, empirically verified)
`npx tsc` / `vite build` fail: `pdf` is referenced out of scope in `web_ui/src/lib/processing/pdf-extractor.ts`'s `finally` block, plus type errors in `xlsx-extractor.ts`. Any "phases done" status in project docs is currently unshippable.
**Fix:** Resolve both compile errors. Add a CI job that runs `tsc --noEmit` and `vite build` on every PR so this cannot regress silently again.

### 2. No usable model weights exist anywhere in the repo (CRITICAL, empirically verified)
`models/bge-small-en-v1.5/onnx/model.onnx` is a 134-byte Git-LFS pointer stub (not the real weights); reranker and LLM GGUF weights are absent entirely. `npm run prepare-models` therefore cannot succeed from a fresh checkout, and no offline archive built from this repo can have working embeddings or chat.
**Fix:** Restore real weights via `git lfs pull` (or equivalent artifact storage — LFS may not be the right long-term answer for multi-GB GGUFs; use judgment) and make `prepare-models.mjs` **fail loudly and non-zero-exit** if a source file is an LFS pointer stub (detect via the `version https://git-lfs.github.com/spec/v1` header) instead of silently copying garbage into `public/models/`.

### 3. SPA fallback defeats every model-presence probe (adjusted CRITICAL→HIGH, reproduced independently)
`web_ui/src/lib/models/model-manifest.ts`'s `checkPackagedModels()` and every other presence probe (`WllamaService.doInitialize`, `ModelReadinessGate.checkModelCached`) treat `res.ok` from a `HEAD`/`fetch` as "file present." But `vite dev` **and** `vite preview` both serve `index.html` with HTTP 200 for any unmatched path (SPA fallback), so a build with zero model files still reports "Packaged Models: Ready" in Settings and passes PACKAGING.md §4's validation procedure.
**Fix:** Don't trust `res.ok` alone. Either (a) verify `Content-Length` against an expected size recorded in the manifest, (b) verify `Content-Type` isn't `text/html`, or (c) read the first bytes and check magic numbers (GGUF: `0x47475546`; ONNX: valid protobuf header) — pick the cheapest option that reliably distinguishes "real model file" from "index.html fallback." Update `public/models/manifest.json` to carry expected sizes if you go that route.

### 4. Test suite is rotted (empirically verified)
Default `npx vitest run` crashes with a heap OOM. With a 12 GB forks pool, 21 files / 76 tests fail — the majority trace to `vitest.config.ts:10` (`setupFiles: []`), so `@testing-library/jest-dom` matchers (`toBeInTheDocument` etc.) are never registered, producing `Invalid Chai property: toBeInTheDocument` across many component test files.
**Fix:** Add the jest-dom setup file to `vitest.config.ts`'s `setupFiles`, tune pool/memory settings so the suite runs without a heap-size workaround, and get the suite green (or explicitly skip/quarantine tests that fail for reasons unrelated to this PR's scope, with a comment explaining why — don't silently delete coverage). Wire `vitest run` into CI once green.

### 5. `validate-build` and `public/models/manifest.json` omit the browser-LLM runtime (HIGH, verified)
`public/models/manifest.json` and the `validate-build` script don't list the wllama WASM runtime, the offline-compat runtime (`@wllama/wllama-compat`), or the LLM GGUF+mmproj files as required. A build can pass packaging validation while shipping an archive that cannot run browser-local chat at all.
**Fix:** Add these to the manifest (mark required for the default `wllama` engine) and to whatever script `validate-build`/`prepare-models` run, so a build missing them fails CI instead of failing silently at runtime.

### 6. Root-absolute `/models` paths break any subpath deployment (HIGH)
`web_ui/src/lib/models/model-manifest.ts:23` hardcodes `MODELS_BASE = '/models'`, but `vite.config.ts:64` deliberately sets `base: './'` so the built archive works when served from any path (e.g. `https://host/training/`, a common IIS/static-host reality). Served from a subpath, every model fetch 404s at the origin root.
**Fix:** Derive `MODELS_BASE` from `import.meta.env.BASE_URL` (or `new URL('models/…', document.baseURI).pathname`) instead of a hardcoded leading slash.

### 7. Production build ships sourcemaps (LOW)
`vite.config.ts:100` sets `sourcemap: true` unconditionally — archive bloat and source exposure for what's meant to be a STIG-scannable offline artifact.
**Fix:** Disable sourcemaps for production builds (or gate behind an env flag for internal debug builds only).

### 8. `public/models/manifest.json` has drifted from `model-manifest.ts` (LOW/INFO)
The committed JSON manifest doesn't match the TypeScript source of truth (missing entries per #5 above) and nothing actually reads the JSON file at runtime — it may be entirely vestigial.
**Fix:** Either delete it if truly unused (grep first to confirm), or generate it from `model-manifest.ts` at build time so the two can't drift again.

## Acceptance criteria
- [ ] `npm ci && npx tsc --noEmit` exits 0
- [ ] `npm run build` succeeds and produces `dist/`
- [ ] `npx vitest run` completes without OOM and without jest-dom matcher errors (pre-existing unrelated failures may remain but must be explicitly noted, not silently fixed by deleting tests)
- [ ] `npm run prepare-models` either succeeds with real weights present or fails loudly with a clear message when they're LFS stubs
- [ ] `checkPackagedModels()` returns `allReady: false` against a `vite preview` server serving **no** model files (currently returns `true` — this is the regression test for finding #3)
- [ ] A build served from a non-root subpath (e.g. via a reverse-proxy prefix) successfully loads model files
- [ ] CI runs typecheck + build + test on every push to this branch

## Do NOT do in this PR
Application logic fixes (chat init, RAG grounding, document ingestion, settings, UI/theming) belong to sibling PRs — see the tracking list below. Keep this PR to build/tooling/packaging only so it merges fast and unblocks everyone else.

## Related
- Full audit: `WEB_UI_AUDIT_REPORT.md` on `claude/html-app-audit-rag-1uszpr`
- Sibling PRs: #21 (chat engine), #22 (RAG retrieval), #23 (document ingestion), #24 (settings), #25 (UI/a11y), #26 (safety epic)
- Related prior audits: #14 (original HTML5 RFC), #18 (UI/UX look-and-feel audit — overlaps with this PR only on manifest drift, no conflict)

zaxbysauce:
Closes #20

## Summary

Resolves #20 — the hard blocker from the `web_ui/` audit (PR 1 of 7). This PR makes the web app **compile, package, and self-check honestly** so the other 6 PRs (chat engine, RAG retrieval, document ingestion, settings, UI/a11y, safety) can be built and tested against a trustworthy foundation. Scope is strictly build / tooling / packaging; application logic owned by sibling PRs is untouched (a small number of unavoidable type errors are deferred via scoped `@ts-expect-error` directives and `as` casts, each naming the owning sibling PR).

Closes #20.

## What changed (by finding)

- **#1 Compile errors (CRITICAL):** Fixed the block-scope bug in `pdf-extractor.ts` (`let pdf` was invisible to the outer `finally`, causing TS2304 + a latent `undefined.destroy()` runtime crash) and the wrong SheetJS generic in `xlsx-extractor.ts` (`Record<string,unknown>` → `unknown[][]` for `header: 1`). Added regression tests for both.
- **#2 LFS-stub weights (CRITICAL):** `prepare-models.mjs` now detects the `version https://git-lfs.github.com/spec/v1` header (via shared `scripts/lib/lfs-detect.mjs`) and **fails loudly, non-zero** on a pointer stub — covering the case a size-only check misses (a >1KB padded pointer). Real weights restored via `git lfs pull`.
- **#3 SPA-fallback false-positive (HIGH):** New `src/lib/models/probe.ts` treats `200 + Content-Type: text/html` as "not present" (the Vite SPA-fallback signature). Wired into all 3 probe sites: `model-manifest.ts` `checkPackagedModels()`, `wllama-service.ts` `isPresent()`, and `model-readiness.ts` `isModelAvailable()`. A build serving zero model files now correctly reports `allReady: false`.
- **#4 Test suite rotted (CRITICAL):** Centralized jest-dom setup (`src/test/setup.ts`), tuned the vitest pool (`forks`, bounded `maxForks`, per-fork `memoryLimit: '3GB'` as a string unit — a bare number is treated as raw bytes by tinypool, so the original `3072` meant ~3KB, not 3GB; now corrected, and per-file `isolate: true` makes the limit a backstop), and quarantined the pre-existing unrelated failures. Each entry in `vitest.config.ts`'s `test.exclude` is annotated with its owning sibling PR (#21–#25) and failure category (none deleted). (Pre-existing `.skip()` calls elsewhere in the suite predate this PR and are not annotated by it.) Suite: **44 files, 788 passing, 1 skipped, 0 OOM**.
- **#5 Missing browser-LLM runtime in manifest (HIGH):** `public/models/manifest.json` now declares the wllama WASM/compat runtime + LFM2-VL GGUF/mmproj, each tagged with a packaging `group` (`core`/`optional`/`llm`). `validate-build.mjs` gained a `--no-llm` opt-out for embeddings-only builds. `prepare-models --no-llm` writes `VITE_EXCLUDE_MODEL_GROUPS=llm` so the runtime readiness gate (`checkPackagedModels`) reports the `llm` group as `excluded: true / ready: true` — an embeddings-only/server-mode build no longer ships with a permanently-broken "models not ready" UI.
- **#6 Root-absolute `/models` breaks subpath deploy (HIGH):** `MODELS_BASE` is now an absolute, deploy-aware path derived from `import.meta.env.BASE_URL` against `document.baseURI` — `/models` at origin root, `/training/models` under a subpath, stable across client-side routing.
- **#7 Production sourcemaps (LOW):** `vite.config.ts` now uses `sourcemap: command === 'serve'` (dev only); the offline artifact no longer ships `.map` files.
- **#8 manifest.json drift (LOW):** Made `public/models/manifest.json` the single source of truth; `model-manifest.ts` imports it at runtime (Vite JSON import) so the TS readiness gate and the build validator cannot drift.
- **CI:** New `.github/workflows/web-ui.yml` runs typecheck + build + test + prepare-models + validate-build on every push/PR. (Previously the entire 60-file vitest suite had **no CI** — the structural cause of the rot.)

## Scope reconciliation

The issue asked for both "`tsc --noEmit` exits 0" **and** "do not do application-logic fixes." Under the single-tsconfig layout those conflicted (247 tsc errors). Resolved by: (1) bumping `target`/`lib` ES2020→ES2022 (kills 12 `{cause}` errors, zero edits), (2) adding `@webgpu/types` (kills 10 WebGPU errors, zero edits), (3) splitting tsconfig so bare `tsc --noEmit` type-checks app source, (4) ~15 genuinely type-only patches to out-of-scope source (drop unused imports, widen a pipeline type, add one missing import), and (5) 2 errors deferred via explicit `@ts-expect-error` directives naming the owning sibling PR (dead `dispose()` cleanup → #21) plus a few `as` casts for API-type drift (flexsearch union, wllama `requestAdapterInfo`) → #22. **No pre-existing application logic (chat, RAG, inference, search, settings) was changed.** A small amount of new runtime logic was added in `model-manifest.ts` (the deploy-aware `MODELS_BASE` derivation, manifest-driven model derivation, and the `--no-llm` exclusion signal) as part of the packaging-integrity fixes themselves.

## Invariant audit

No `AGENTS.md`/`CONTRIBUTING.md` invariants; defaults applied and verified:
- **Build succeeds:** `npm run build` → exit 0, `dist/` produced (`tsc -b && vite build`).
- **Typecheck passes:** `npx tsc --noEmit` → exit 0 (app source); `npx tsc --noEmit -p tsconfig.test.json` → exit 0 (full test tree).
- **Tests pass:** `npm test` → 44 files / 788 passed / 1 skipped / 0 failed / no OOM.
- **No generated artifacts staged:** `.zcode/` working state gitignored; `web_ui/node_modules` and `dist/` already ignored. (`web_ui/tsconfig.tsbuildinfo` was tracked before this PR and is regenerated — left as-is to avoid scope creep.)
- **No secrets/PII:** pre-push secret scan clean (no `sk_live_`/`ghp_`/`AKIA`/JWT/etc.).
- **Conventions:** conventional-commit title `fix(web_ui): ...` matches repo history; CHANGELOG `[Unreleased]` entries added; no release-fragment convention in this repo.

## Acceptance criteria

- [x] `npm ci && npx tsc --noEmit` exits 0
- [x] `npm run build` succeeds and produces `dist/` (no sourcemaps)
- [x] `npx vitest run` completes without OOM and without jest-dom matcher errors (44 files, 788 passing; pre-existing failures explicitly quarantined with owning-PR notes)
- [x] `npm run prepare-models` succeeds with real weights, fails loudly on LFS stub (size + header detection)
- [x] `checkPackagedModels()` returns `allReady: false` against SPA fallback (regression test in `probe.test.ts` / `model-manifest.test.ts`)
- [x] Build served from a subpath loads model files (deploy-aware absolute `MODELS_BASE`)
- [x] CI runs typecheck + build + test on every push (`.github/workflows/web-ui.yml`)

## Test plan

- `npx tsc --noEmit` → exit 0
- `npm run build` → exit 0, `dist/` produced, no `.map` files
- `npm test` → 44 files, 788 passing, no OOM
- `npm run prepare-models` (with real weights) → exit 0
- `node scripts/validate-build.mjs --no-llm` (after prepare + build) → exit 0
- New unit tests: `probe.test.ts` (8), `model-manifest.test.ts` regression (2), `pdf-extractor.test.ts` cleanup (2), `lfs-detect.test.ts` (6)

## Notes for reviewers / sibling PRs

- **Sibling PR #21 (chat engine):** `useServiceInitialization.ts:267,269` has two `@ts-expect-error` directives guarding a `readinessGate.dispose()` call that doesn't yet exist on `ModelReadinessGate` (runtime-guarded with `typeof === 'function'`, so it's a no-op today). **When you add `dispose()` to `ModelReadinessGate`, remove those two directives or TS2578 will fire.** Also `wllama-service.ts` / `model-readiness.ts` were edited to consume the hardened probe (finding #3) — adopt `probeAsset` from `src/lib/models/probe.ts`.
- **Sibling PR #22 (RAG retrieval):** `keyword-index.ts:218` flexsearch `search()` union cast + `keyword-index.ts:112` `worker` option cast are scoped casts awaiting your fix.
- **Sibling PR #23 (document ingestion):** the extractor test files (`docx/xlsx/pptx-extractor.test.ts`) are stubs that assert a stale `string` return shape; quarantined, awaiting your real tests.
- **Sibling PR #25 (UI/a11y):** component tests with stale import paths were fixed (off-by-one directory level); some remain quarantined for runtime issues.
- **Acceptance #5 note:** the SPA-fallback regression for `checkPackagedModels()` is covered by a mocked-fetcher unit test (`probe.test.ts`, `model-manifest.test.ts`) that faithfully reproduces Vite's exact response signature (HTTP 200 + `Content-Type: text/html`), rather than a live `vite preview` integration test. Functional coverage is equivalent; a live-server test can be added if a maintainer prefers the literal wording.
- **Sibling PR heads-up (security lane):** `api/client.ts` changed `getEffectiveToken()` from `this.token || getToken()` to `this.token ?? getToken()`. For an empty-string `this.token`, `??` no longer falls through to the stored token (empty string is not nullish). No call site constructs `ApiClient` with an empty-string token today (the singleton uses `''` as `baseUrl`, not token), and all consumers gate on `if (token)` where `''` is falsy — so there is no observable difference. Flagging for #21/#24 in case a future call site relies on the empty-string fallthrough.

## Review feedback addressed

Addressed in a follow-up commit after review:
- **F1 (HIGH):** `model-manifest.ts` was silently dropping the manifest `group` field, so `checkPackagedModels()` had no way to know a build was `--no-llm` and would report the `llm`-group files as missing forever (`allReady: false`) on a valid embeddings-only/server-mode deployment. Fixed: `group` is now threaded through `PackagedModel`/`ModelReadiness`; `prepare-models --no-llm` writes `VITE_EXCLUDE_MODEL_GROUPS=llm` into `.env.production`; `checkPackagedModels` reports excluded-group models as `excluded: true / ready: true` and does not probe or count their files as missing. Regression tests added.
- **F2 (MEDIUM):** `vitest.config.ts` `memoryLimit: 3072` was ~3KB, not 3GB (tinypool treats a bare number as raw bytes; the `isolate: true` flag made the limit unreachable, which is why the suite still passed). Corrected to `memoryLimit: '3GB'` so the limit is real if `isolate` is ever removed.
- **F3 (LOW, doc):** "No application logic changed" was overclaimed — `model-manifest.ts` adds real new runtime logic (`resolveAbsoluteBase`, manifest-driven derivation, exclusion signal). Corrected in the scope-reconciliation section above and in CHANGELOG.md.
- **F4 (LOW, doc):** "each quarantined test annotated" was imprecise — it applies to the `test.exclude` entries this PR added; pre-existing `.skip()` calls elsewhere predate this PR. Refined in finding #4 above.
- **P3 (cosmetic):** removed the stale "Status: SKIP - vitest not installed" header from `pdf-extractor.test.ts`.
- **P2 (CI cost):** switched `web-ui.yml` from `windows-latest` to `ubuntu-latest` (the Python jobs need Windows; this pure-Node/Vite job does not).
- **P1 (supply-chain, not addressed here):** no `permissions:` block and actions pinned to tags not SHAs — standard hardening, not specific to this PR's risk profile (no `pull_request_target`, no untrusted-input interpolation). Left for a repo-wide CI-hardening pass.


zaxbysauce:
Thanks for the review — all confirmed findings addressed in `1c9d069`. Closure ledger below.

| id | severity | item | outcome | fix-ref |
|----|----------|------|---------|---------|
| F1 | HIGH | `model-manifest.ts` dropped `group`; `checkPackagedModels` reported llm-group files missing forever on `--no-llm` builds | **FIXED** | `model-manifest.ts`: `group` now on `PackagedModel`/`ModelReadiness`; `EXCLUDED_MODEL_GROUPS` reads `VITE_EXCLUDE_MODEL_GROUPS` (written by `prepare-models --no-llm`); excluded models → `ready:true, excluded:true`, not probed, not in `missing`. 5 regression tests. |
| F2 | MEDIUM | `vitest.config.ts` `memoryLimit: 3072` was bytes (~3KB), not MB | **FIXED** | `memoryLimit: '3GB'` (string). Comment notes it's a latent backstop — the forks pool doesn't forward it and `isolate:true` short-circuits the recycle check; the real OOM protection is `maxForks:2` + fresh-fork-per-file. |
| F3 | LOW (doc) | "No application logic changed" overclaimed | **FIXED** | PR body scope-reconciliation + CHANGELOG now explicitly note the new runtime logic in `model-manifest.ts` (`resolveAbsoluteBase`, manifest derivation, exclusion signal); "no logic changed" scoped to chat/RAG/inference/search/settings. |
| F4 | LOW (doc) | "each quarantined test annotated" imprecise | **FIXED** | Refined to "each `test.exclude` entry annotated" (the 18 this PR added); pre-existing `.skip()` calls elsewhere predate this PR. |
| P1 | pre-existing | no `permissions:` block; actions pinned to tags not SHAs | **NOT ADDRESSED** | Standard repo-wide CI hardening; no `pull_request_target` or untrusted-input interpolation here. Left for a repo-wide pass. |
| P2 | pre-existing | `web-ui.yml` on `windows-latest` for a Node pipeline | **FIXED** | → `ubuntu-latest` (Python jobs need Windows; this job does not). |
| P3 | pre-existing | stale "Status: SKIP" header in `pdf-extractor.test.ts` | **FIXED** | Replaced with an accurate description. |
| P4 | pre-existing | `client.ts` `||`→`??` empty-string-token behavior change | **DOCUMENTED** | Flagged in the PR body for sibling PRs #21/#24 — no call site constructs with an empty-string token today, and all consumers gate on `if(token)` where `''` is falsy, so no observable difference. |

**Validation (re-run after fixes):** `tsc --noEmit` exit 0 (app + test configs); `npm run build` exit 0, no sourcemaps; `npm test` → 44 files / 788 passing / 1 skipped / no OOM; `prepare-models --no-llm` → writes `VITE_EXCLUDE_MODEL_GROUPS=llm`; `validate-build --no-llm` → exit 0.

On F1's CI coverage (the reviewer noted the runtime path wasn't exercised): the bundle-grep approach was fragile under minification, so I replaced it with (a) a CI step verifying `prepare-models --no-llm` writes the marker to `.env.production`, and (b) 3 minification-independent vitest unit tests that `vi.stubEnv('VITE_EXCLUDE_MODEL_GROUPS', ...)` and assert the `EXCLUDED_MODEL_GROUPS` parsing (positive, negative/empty, and invalid-name-filtering cases). The deeper end-to-end (Vite inlining) was independently verified by inspecting the built bundle.

PR #28, opened 2026-07-13
fix(web_ui): chat engine initialization & session lifecycle (Issue #21)

### LINKED ISSUE #21: [web_ui PR-2/7] Chat Engine Initialization &amp; Session Lifecycle (fixes "chat sometimes doesn't work")

zaxbysauce:
Closes #21

## Summary

PR-2/7 of the `web_ui/` audit. Closes the root causes of the user complaint **"chat sometimes doesn't work"** — 11 findings across browser-LLM initialization, session lifecycle, model readiness, streaming, and the chat UI.

**Root cause of the headline defect (F1):** the browser-local chat path (`runGeneration`) fetched an *uninitialized* LLM singleton from the factory and passed it straight to `RAGOrchestrator.query()`, which calls `generate()` before any `initialize()` — so every send in the default wllama config threw `"WllamaService not initialized. Call initialize() first."` into the assistant bubble.

See the commit messages and the per-finding list below for the full scope. Three commits:
- `5498e42` — original implementation of all 11 findings (F1–F11).
- `a76e7de` — addresses the validated findings from a `swarm-pr-review` pass on this PR (PRR-001…PRR-013).
- `4fe090a` — addresses 22 findings from a second, independent `swarm-pr-review` pass (see "Second review round" below).

### Findings closed (F1–F11)
- **F1 (CRITICAL):** `runGeneration` now `await`s `llmService.initialize(modelId, onProgress)` before constructing the orchestrator. Idempotent — safe on every send.
- **F2 (HIGH):** engine dispose moved out of unmount into a genuine engine-switch effect (also aborts any in-flight generation first — PRR-010).
- **F3 (HIGH):** single source of truth for the WebLLM model id; Cache Storage probe for cross-session readiness; re-dispatch readiness after download.
- **F4 (HIGH):** `getDeviceMemory()` returns `8` on unsupported browsers; per-model requirement for `lfm2-vl-1.6b`.
- **F5 (HIGH):** stream-setup wrapped in try/catch routing to `onError` (no permanent wedge); LAN/private server URLs allowed.
- **F6 (MEDIUM):** SSE consumer synthesizes an error completion on unterminated close.
- **F7/F8 (MEDIUM):** readiness in-flight promise keyed by engine from creation; listener discards stale-engine events.
- **F9 (HIGH):** streaming path passes `getToken()`; `PACKAGING.md` documents opt-in auth.
- **F10 (HIGH):** engine-aware overlay with real failures/recommendations + Retry/Settings.
- **F11 (misc):** WebGPU watchdog wired end-to-end; `adapter.info`; WebLLM `initialize` in-flight guard; single keyboard-shortcut registrar; recoverable RAG errors degrade; boot connectivity; StrictMode-safe deferred cleanup.

### Review-driven additions (commit `a76e7de`, PRR-001…PRR-013)
A `swarm-pr-review` pass (6 explorer lanes → 2 GLM-5.2 reviewers → critic) found 10 validated findings; all are resolved:
- **PRR-002 (blocker, AC6):** Ctrl/Cmd+Enter now sends while the chat input textarea is focused (the global shortcut bails on TEXTAREA targets, so a `handleKeyDown` branch handles it). AC6 met in the primary case.
- **PRR-001:** added `ChatPage.init.test.tsx` proving `initialize()` runs before `query()` on a cold send.
- **PRR-003:** watchdog recovery now dispatches readiness events (block on teardown, re-check after recovery) and surfaces recovery failures to the UI via a `webgpu-recovery-failed` listener.
- **PRR-004:** `strictModeDisposeTimer` moved to module scope so a real unmount + rapid remount cancels the prior instance's deferred dispose.
- **PRR-006:** stale `webgpu-watchdog.test.ts` assertions updated to `WEBLLM_DEFAULT_MODEL_ID`.
- **PRR-007:** `ChatInput.handleClear` syncs `onDraftChange('')` so `draftRef` stays correct after the ✕ button.
- **PRR-010:** engine-switch effect aborts the in-flight generation before disposing the old engine.
- **PRR-011:** boot connectivity surfaces auth-required when the server reports `ENABLE_AUTH=true` and no token is stored.
- **PRR-012/013:** stale "OPFS" comments fixed; model-id single-source-of-truth completed (`checkMemory` default, `model-download` size table, `SettingsPage`).

Rejected/disproved review candidates (no action): PRR-005 (test passes), PRR-008 (marginal impact), PRR-009 (window unreachable), PRR-017 (safe).

### Second review round (commit `4fe090a`, 22 findings)

A second, independent `swarm-pr-review` pass (6 explorer lanes + 2 triggered micro-lanes → 6 independent reviewer chunks → 3 critic passes) found 22 additional validated findings, closed in this commit through a full Stage A / Stage B / closeout-gate cycle:

**Two HIGH-severity ship blockers:**
- **F-LEAK:** the browser engine was never disposed when the engine preference changed via Settings, because `ChatPage`'s local ref got re-seeded to the new value on every remount, permanently skipping disposal (a regression vs. this PR's own earlier `5498e42` behavior). Disposal now lives in `InferenceModeContext`'s `setBrowserEngine`, which survives `ChatPage` unmount/remount and fires regardless of the current inference mode — closing a second leak path found mid-review (changing engine preference while in API mode).
- **F-TEST:** `ChatPage.init.test.tsx`'s regression test for the CRITICAL F1 fix recorded call order synchronously and could not actually detect a missing `await` — verified via an executed mutation probe (removing the `await` still passed the old test). It now crosses a real macrotask boundary and asserts completion state at the moment `query()` is invoked.

**Also closed:** a security-relevant finding (F-SSRF — `validateStreamUrl`'s comment claiming server-side SSRF compensation was factually false, since the server's `validate_url()` is dead code; comment corrected, an IPv6 metadata-address carve-out gap closed, a first-byte fetch timeout added), readiness-gate dedup/state fixes (F-DEDUP, F-WEBLLM-READINESS-GAP), Clear Cache now actually clearing WebLLM's Cache Storage entries instead of silently no-op'ing on OPFS (F-CLEAR-CACHE-OPFS), a restored global Ctrl+, shortcut (F-KEYBOARD-REGRESSION), and several test-coverage and documentation-accuracy fixes (full list of 22 finding IDs in the commit message).

Went through Stage A (typecheck/build/full test suite), Stage B (independent reviewer + test_engineer running executable falsification probes — not just re-reading tests), and a closeout reviewer + critic pass; two closeout rounds found and closed additional test-coverage gaps beyond the initial fixes.

**Explicitly deferred, not silently dropped:** `web-llm-service.ts`'s `_generation` dispose-race guard (F-WATCHDOG) and the `adapter.info` capability check (F-ADAPTER-INFO-COMPAT) have correct, critic-verified fixes but no CI-enforced regression test — both LOW severity, already downgraded once on reachability grounds, and `web-llm-service.test.ts`/`webgpu-watchdog.test.ts` are both in the vitest exclude list. Recommend a small follow-up adding a fake-`GPUAdapter`-based test for both.

## Invariant audit

| Invariant (from CI / web-ui.yml) | Evidence |
|---|---|
| `npx tsc --noEmit` (app source) exits 0 | clean |
| `npx tsc --noEmit -p tsconfig.test.json` exits 0 | clean |
| `npm run build` succeeds, produces dist/ | `✓ built` (only pre-existing chunk-size warning) |
| `npm test` (`npx vitest run`) | **51 files, 815 passed, 1 skipped** (up from 45/793 before the second review round) |
| No generated/build artifacts staged | `git status` shows only source + test files |
| No secrets in diff | pre-push scan clean |
| Conventional commit title | `fix(web_ui): ...` |

## Test plan

```
npm test              → 51 files, 815 passed, 1 skipped
npm run typecheck:all → EXIT 0 (both tsconfigs)
npm run build         → ✓ built
```

### Adversarial review
- Original implementation: independent GLM-5.2 reviewer found 4 bugs (2 blockers: async `requestDevice()`, StrictMode timer); all fixed. Final critic APPROVE.
- First feedback round (`swarm-pr-review`, commit `a76e7de`): 6 explorer lanes → 2 GLM-5.2 reviewers → critic found 10 validated findings; all fixed. Independent reviewer APPROVE; final critic APPROVE (82% → gap closed by wiring the `webgpu-recovery-failed` listener).
- Second feedback round (`swarm-pr-review` + `swarm-pr-feedback`, commit `4fe090a`): 6 explorer lanes + 2 micro-lanes → 6 reviewer chunks → 3 critic passes found 22 validated findings (2 HIGH). Closed via Stage A/B + closeout reviewer/critic gates, two of which found and closed additional test-coverage gaps.

### Caveats
- **Corrected:** several owning test files (`ChatPage.rag/server-mode/shortcuts`, `TokenStreamManager`, `webgpu-watchdog`, `web-llm-service`, `InferenceModeContext`, and — not disclosed in the original version of this note — `ChatPage.test.tsx`/`ChatPage.verification-3.3.test.tsx`/`SettingsPage.test.tsx`) remain in the vitest `exclude` list (pre-existing drift, not introduced by this PR). The second review round closed the highest-value coverage gaps this created by adding new, non-excluded, purpose-built test files instead of relying on the excluded files (`App.test.tsx`, `App.keyboard.test.tsx`, `ChatPage.overlay.test.tsx`, `ChatPage.keyboard-guard.test.tsx`, `InferenceModeContext.disposal.test.tsx`, `SettingsPage.clearCache.test.tsx`). Two low-severity items (F-WATCHDOG, F-ADAPTER-INFO-COMPAT) remain covered only by inspection/ad-hoc falsification probes, not a permanent CI-enforced test — see "Explicitly deferred" above.
- `onOpenSettings` is added to `ChatPageProps`; PR-6 (#25) owns the `renderPage` switch and should rebase.

## Coordination note

PR-6 (#25) owns App.tsx routing. **Corrected from the original version of this note:** this PR's `App.tsx` changes are not a pure one-line passthrough — besides adding the `onOpenSettings` prop into the `'chat'`/`default` switch cases (the `'documents'` case is untouched), the second review round also removed the `useKeyboardShortcuts` import/call from `App.tsx` (F2/F11's registrar consolidation) and later restored a smaller global registration for the Ctrl+, shortcut only (F-KEYBOARD-REGRESSION). PR-6 (#25) should rebase onto this PR's final `App.tsx` state, not assume only the `onOpenSettings` prop changed.


zaxbysauce:
## swarm-pr-feedback closure ledger

Round: second independent `swarm-pr-review` → `swarm-pr-feedback` cycle on this PR. Fixes pushed at `4fe090a`.

**Intake sources checked:** PR body/checkboxes, linked issue #21 acceptance criteria, commit history, CI checks (5/5 green pre-round), review comments/threads (none existed pre-round — this was the review that produced the findings below). No sources were unavailable.

**Ledger counts:** 22 CONFIRMED and fixed · 1 DISPROVED (no action) · 2 PRE_EXISTING (no action, unrelated to this PR) · 2 PR-body-accuracy items (fixed via PR description update, not code).

| ID | Severity | Status | Evidence |
|---|---|---|---|
| F-LEAK | HIGH | **fixed** | Disposal moved to `InferenceModeContext.setBrowserEngine`, mode-agnostic; `InferenceModeContext.disposal.test.tsx` + Stage B executable mutation probe (reverted fix → test fails, restored → passes) |
| F-TEST | HIGH | **fixed** | `ChatPage.init.test.tsx` rewritten to assert completion-at-invocation-time, not call order; verified twice via executed mutation probes (Stage B test_engineer + closeout critic) |
| F-SSRF | MEDIUM | **fixed** | Corrected false "server compensates" comment in `streaming.ts` (server's `validate_url()` confirmed dead code) |
| F-IPV6-METADATA-GAP | LOW | **fixed** | `isIpv6MetadataAddress()` added; verified via mutation probe |
| F-NO-FETCH-TIMEOUT | LOW | **fixed** | 30s first-byte timeout, cleared on first data/close, distinguishes auto-timeout from user cancel |
| F5-04-FLAKY-HELPER | LOW | **fixed** | `flushMicrotasks(times)` now honors its parameter |
| F-AC3 | MEDIUM | **fixed** | 3 new tests exercise the Cache Storage cross-session probe via `vi.stubGlobal` |
| F-AC7 | MEDIUM | **fixed** | New `ChatPage.overlay.test.tsx` (5 tests) covers the blocked-overlay render |
| F-MISSING-NEGATIVE-TEST | MEDIUM | **fixed** | Added rejecting-`initialize()` test in `ChatPage.init.test.tsx` |
| F-KEYBOARD-REGRESSION | MEDIUM | **fixed** | Global Ctrl+, restored in `App.tsx`; covered by new `App.keyboard.test.tsx` |
| F-STALE-TEST-DOC | LOW | **fixed** | Docstring/assertions corrected in `ChatPage.shortcuts.test.tsx` (note: this file itself remains vitest-excluded — real CI coverage lives in the new `App.keyboard.test.tsx`) |
| F-AC2-COVERAGE-GAP | MEDIUM | **fixed** | New `App.test.tsx` (2 tests) proves conversation state survives page navigation |
| F-CLEAR-CACHE-OPFS | MEDIUM | **fixed** | Clear Cache now also deletes the 3 Cache Storage caches WebLLM actually uses; new `SettingsPage.clearCache.test.tsx` |
| F-STALE-OPFS-COMMENT | LOW | **fixed** | Comment corrected to "Cache Storage" |
| F-WATCHDOG | LOW | **fixed** (test **deferred**) | `_generation` counter guard in `web-llm-service.ts`; correctness proven via an executed Stage B falsification probe, but no permanent CI-enforced test landed (see below) |
| F-DEDUP | LOW | **fixed** | `finally` block now checks promise identity before clearing; new dedup regression test |
| F-WEBLLM-READINESS-GAP | LOW | **fixed** | `modelReady = ready && modelCached` now applies uniformly; test added |
| F-CTRL-ENTER-BYPASS | LOW | **fixed** | Global Ctrl+Enter now checks `!isInputDisabled`; new `ChatPage.keyboard-guard.test.tsx` |
| F-TOKEN-KEY-DUP | LOW | **fixed** | Uses `getToken()` instead of a duplicated literal |
| F-AUTH-MSG-DOCS-MISMATCH | LOW | **fixed** | Error message reworded to match actual (operator-only) recovery path |
| F-ADAPTER-INFO-COMPAT | LOW | **fixed** (test **deferred**) | Explicit capability check + `console.warn`; no permanent CI-enforced test (see below) |
| F-CHANGELOG-GAP | LOW | **fixed** | `CHANGELOG.md` entry added and iteratively corrected across 2 closeout rounds |
| F-EXCLUDE-LIST-CLAIM | MEDIUM | **fixed (PR body)** | PR description's exclude-list caveat corrected to disclose the 2 previously-undisclosed excluded files |
| F-PR-CLAIM-APP-TSX | LOW-MEDIUM | **fixed (PR body)** | Coordination note corrected — `App.tsx` changes are not a one-line passthrough |
| F-EFFECT-THRASH | — | disproved | Pre-existing render-thrashing pattern, not introduced by this PR |
| F5-04-TAUTOLOGY | — | pre-existing | Doubly inert (skipped + vitest-excluded), unmodified by this PR |
| F5-06-MEMORY-COVERAGE | — | disproved | A pre-existing test already covers the claimed gap |

**Explicitly deferred (disclosed, not silently dropped):** F-WATCHDOG and F-ADAPTER-INFO-COMPAT have correct, critic-verified production fixes but no *permanent* CI-enforced regression test — both LOW severity (already downgraded once on reachability grounds during the original review), and the natural home for such a test (`web-llm-service.test.ts`/`webgpu-watchdog.test.ts`) is itself vitest-excluded for unrelated pre-existing reasons. A proper test needs a controllable fake `GPUAdapter`. Recommend a small follow-up.

**Validation:** Stage A (typecheck ×2, build, full suite) → green. Stage B: independent reviewer (found and the round fixed 1 residual gap) + test_engineer (6 executed falsification probes, all PASS). Closeout gate: independent reviewer + final critic, two rounds, each finding and closing additional test-coverage gaps before final APPROVE-equivalent state. Final suite: **51 files / 815 passed / 1 skipped** (up from 45/793).

**Review-thread resolution:** not performed — no review threads existed on this PR before this round (0 reviews, 0 review comments), and none were created since findings came from an independent `swarm-pr-review` run rather than inline PR comments.

**Mergeability:** `mergeable_state` is `blocked` with no CI failures and no merge conflicts detected — appears to reflect a required-approval branch-protection rule rather than a code or CI blocker; unaffected by this round's changes.

Full finding ledger with provenance (explorer → reviewer → critic) and gate history: `.swarm/pr-review/pr-28-20260714084110/findings.jsonl` (review run) and this session's local gate records.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #37, opened 2026-07-16
[PR-B] Retrieval engine & air-gapped delivery: resurrect the reranker, fix relevance gating/chunking/keyword search, harden offline packaging (web_ui)


zaxbysauce:
# PR-B — Retrieval engine & air-gapped delivery

## Context for the implementer

`web_ui/` is the sole release artifact: **one HTML5 app, fully air-gapped (zero runtime network), 12th-gen Intel i5, no dedicated GPU**. Retrieval stack: transformers.js (`@huggingface/transformers` 3.8.1 pinned) with bge-small-en-v1.5, edgevec 0.6.0 HNSW (cosine), flexsearch 0.8.212 keyword index, RRF fusion, ms-marco-MiniLM-L-6-v2 cross-encoder (intended), wllama LFM2-VL-1.6B GGUF (n_ctx 4096). The owner's complaint: **poor retrieval quality**.

This issue is one of two derived from a multi-lane audit with independent adversarial re-derivation of every claim from pinned library sources at HEAD `41923f5` (including an executed tokenizer reproduction). Sibling: **[PR-A]** (chat correctness & UX) — parallel-implementable; seams listed in §6.

**There are no production users.** Vector re-indexes are free: bump `VECTOR_INDEX_VERSION`, wipe, re-add dev corpora. No migration machinery is required (see §4 for the simplifications this allows).

**⚠ Correction to #35:** #35 listed "real cross-encoder reranker present — verified correct, do not break." That is **wrong** (it verified the file exists, not that it is reachable), and #35's own top recommendation (B1 over-fetch-then-rerank) presupposes a working reranker. Findings R1–R3 below are the actual state.

Line numbers verified at HEAD `41923f5`; if drifted, re-locate by the quoted code.

---

### R1 (CRITICAL): the reranker never runs — and wiring the init alone will make things WORSE

Three stacked defects; **all three must land together**:

[...]

PR #38, opened 2026-07-17
fix(web_ui): chat correctness & UX overhaul — renderer, streaming/persistence integrity, a11y (Issue #36)

### LINKED ISSUE #36: [PR-A] Chat correctness & UX overhaul: fix ordered-list rendering, streaming/persistence data loss, and UX/a11y defects (web_ui)
# PR-A — Chat correctness & UX overhaul

## Context for the implementer

`web_ui/` is the sole release artifact: **one HTML5 app, fully air-gapped (zero runtime network), running in-browser on 12th-gen Intel i5 CPUs with no dedicated GPU**. Stack: React 18 + Vite + TS, wllama (llama.cpp WASM) LLM, transformers.js embeddings, Dexie/IndexedDB persistence. Generations take **minutes** on target hardware — interruption is the common case, not the edge case, which raises the severity of every persistence/cancel bug below.

This issue is one of two implementation issues derived from a multi-lane audit (5 parallel explorers + independent adversarial re-derivation of every high-impact claim from pinned library sources at HEAD `41923f5`). It supersedes the renderer/interaction parts of #35 (Clusters A and C), whose findings were re-validated with several corrections noted inline. The sibling issue **[PR-B]** covers the retrieval engine and air-gapped packaging; the two are designed to be implementable in parallel with only trivial seams (noted in §5).

All line numbers were verified at HEAD `41923f5`. If they have drifted, re-locate by the quoted code, not the number. **This PR must not require a vector re-index** — there are no production users, but keeping this PR frontend-only is what allows it to ship independently of PR-B.

Scope: everything is under `web_ui/src/`.

---

### A1 (CRITICAL, user-reported): ordered lists render every item as "1."

**Root cause** (`components/MarkdownRenderer.tsx`): the list-grouping loop terminates at the first non-matching line — **including blank lines**:

```ts
const ORDERED_RE = /^\d+\.\s+/;                                  // :169
...
while (i < lines.length && ORDERED_RE.test(lines[i].trim())) {   // :220-227
  items.push(lines[i].trim().replace(ORDERED_RE, ''));           // number DISCARDED (:223)
  i++;
}
```

Chat models overwhelmingly emit "loose" lists (`1. a\n\n2. b`), so each item becomes its **own single-item `<ol>`**; the marker value is destroyed by `.replace` and the `<ol>` at `:356-367` carries no `start` attribute → every item renders "1.". Symmetric issue for `UNORDERED_RE` at `:209-217`.

**The "flash" mechanism (corrects #35's explanation):** `ORDERED_RE` requires whitespace after the dot. Mid-stream, the partial line `"2."` (space not yet arrived) fails the regex and renders as a **plain paragraph showing the literal "2."** for a frame; when the next token completes the marker, the line becomes a fresh single-item `<ol>` and collapses to "1.". The collapse happens **item-by-item during streaming** at the flush where each marker gains its trailing space — there is no single "final re-parse" moment (#35's contiguous-buffer-vs-final-text story is impossible for append-only content; content is byte-identical from wllama through persistence — verified `wllama-service.ts:221-224` → `rag-orchestrator.ts:421-427` → `ChatPage.tsx:171-178`, no normalization anywhere).

### Strategy decision (make this first)

**Option B (recommended): replace the hand-rolled parser with `react-markdown` + `remark-gfm`.** Both are pure-JS, fully bundleable, air-gap-safe (no runtime fetches). This closes A1–A5, A8, N2, N3, N9, N11, N12 below in one move. Requirements:
- Do **not** add `rehype-raw` (keeps raw-HTML injection impossible — current code has zero `dangerouslySetInnerHTML`; preserve that property).
- Keep the URL scheme allowlist via `urlTransform` (current allowlist behavior + tests at `MarkdownRenderer.test.tsx:210-293`; also fix N10 below while at it: reject relative and scheme-relative URLs — current `isValidUrl` at `:15-22` resolves against a placeholder base so `//evil.com/x` and `/settings` become live links).
- Map `components` onto the existing design tokens; this is also where A8 (code-block header: language chip + per-block copy button — copy pattern exists at `ChatMessageBubble.tsx:30-44`) lands naturally.
- Bundle growth ~50–70 kB gzipped — irrelevant next to packaged model weights.
- **Streaming perf:** re-parse per token flush already happens today (`useMemo(() => parseMarkdown(content), [content])` at `:394-395`); remark is heavier. Throttle the parse while `message.isStreaming` (parse at most every ~100 ms; always parse final content). This also closes A6.
- Rewrite `MarkdownRenderer.test.tsx` to assert **semantic output** (`<ol start>`, `<li>` counts, `<table>` presence), not text presence. Note the existing test at `:116-122` **asserts broken behavior** (name says "renders bold with __" while asserting literal underscores render) — it must be replaced, not preserved.

**Option A (fallback if B is rejected): patch the parser.** Minimum correct fix: in both grouping loops, peek past runs of blank lines and continue the list block if the next non-blank line matches the same marker family (loose-list semantics); capture the first marker's value and emit `<ol start={n}>`; apply CommonMark's paragraph-interruption rule (only `1.` may interrupt a paragraph — this also fixes N2). ~30–50 lines. The rest of the defect class (below) must then be patched piecewise.

### Renderer defect inventory (all confirmed; closed automatically by Option B, individually by Option A)

| ID | Defect | Evidence |
|---|---|---|
| N2 (MED-HIGH) | Any line starting `<digits>. ` is eaten as a list marker and **the number is deleted**: `"1969. It was a great year"` renders as `"1. It was a great year"` — meaning destroyed | `ORDERED_RE` `:169`, paragraph collector breaks on `isListMarker` `:249-256`, marker stripped `:223` |
| N3 (MED) | Contiguous lists starting at N≠1 renumber from 1 (`"3. a\n4. b"` → "1., 2.") | `:223`, no `start` at `:356-359` |
| A2 (HIGH) | No GFM tables, no `---` hr, no nested lists (indentation discarded by `.trim()` before match and store) | `:212`, `:223`, block union `:156-162` |
| A3 (MED) | `__bold__` / `_italic_` silently unstyled (asterisk-only regex) | `:47`; broken test `:116-122` |
| A4 (MED) | `parseLinks` runs before inline-code tokenization → `` `[a](https://b)` `` becomes a live link | `:101-120` |
| A5 (MED) | Unclosed code fence swallows the rest of the message to EOF (realistic: `maxTokens` truncation mid-fence, or Stop pressed while a fence is open) | `:185-196` |
| A6 (MED) | Full re-parse per token flush; fence-open/list-split reinterpretation flicker. (Bounded: index-keyed earlier blocks are prop-identical so React doesn't touch their DOM — don't over-claim this one) | `:394-395` |
| A7 (LOW) | Empty assistant message renders a bordered box with only a blinking cursor; Stop-before-first-token leaves it permanently, with a Copy button that copies `""` | `ChatMessageBubble.tsx:186-195,220` |
| A8 (LOW) | Code blocks: `block.language` captured (`:186,195`) but never rendered; no copy button | render case `:321-337` |
| N9 (LOW) | Nested emphasis garbles (`**a *b* c**`, `***x***`) | `:47` |
| N10 (LOW) | `isValidUrl` accepts relative/scheme-relative URLs (resolved against a placeholder base). Not XSS (`javascript:`/`data:` blocked) but `[click](anything)` becomes a live link | `:15-22`, href `:109` |
| N11 (LOW) | Catalogued gaps: task lists render literal `[ ]`, `~~strike~~` literal, autolinks/bare URLs not linked, HTML entities undecoded, `1)` markers, setext headings, 4-space code blocks, `~~~` fences | parseBlocks throughout |
| N12 (LOW) | `## Title ##` keeps trailing `##`; blockquote lazy continuation splits | `:200-206`, `:231-239` |

**Renderer facts that are already fine (do not "fix"):** zero `dangerouslySetInnerHTML` in the tree; links carry `target="_blank" rel="noopener noreferrer"` (`:110-111`); dangerous schemes blocked with tests; headings and blockquotes **do** exist (added post-#35 — don't re-add).

---

### S1 (CRITICAL — silent data corruption): switching conversations mid-stream overwrites persisted history

Adversarially verified end-to-end. The mechanism:

1. The only cancel guard fires on a **non-empty→empty** transition: `ChatPage.tsx:434-441` — `if (prev > 0 && messages.length === 0 && tokenStreamManagerRef.current) cancelActiveStream()`. `selectConversation` (`hooks/useConversations.ts:145-157`) swaps to conversation B's **non-empty** array → no cancel. (Saved conversations always have ≥1 message — `saveMessages` early-returns on empty at `:180` — so the guard can never trip on a switch.)
2. No epoch/conversation-id guard exists anywhere in the streaming path. The `tokenStreamManagerRef.current !== streamManager` checks (`:270,:283`) don't help — nothing a conversation switch does clears that ref.
3. `onDone` (`ChatPage.tsx:184-216`) holds a **stale closure**: `saveMessages` is `useCallback(..., [currentConversationId, ...])` (`useConversations.ts:175-217`) reading `currentConversationId` from closure, not a ref; `onDone` captured `onSaveConversation` at send time.
4. `messagesRef` syncs to the incoming prop (`:79-81`), so at completion it contains **B's messages**; the map over `assistantMessageId` matches nothing; `onSaveConversation(updated /* = B's messages */)` runs unconditionally at `:211`.
5. Outcome: A saved → `updateConversation(A_id, { messages: B's messages })` — **A's history replaced with a copy of B's**. A unsaved → a duplicate conversation containing B's messages is created, **and** the stale closure then calls `setCurrentConversationId(newConv.id)` (`useConversations.ts:203`) — re-pointing the active conversation at the duplicate so **every subsequent send forks B's history**.

**Fix (both layers):**
- Cancel the active stream on **any** conversation switch (invoke the same `cancelActiveStream` from `selectConversation`/watch `currentConversationId` in ChatPage), and
- Make saves unable to mis-target: pass the owning `conversationId` into `runGeneration` and thread it through to `saveMessages(conversationId, messages, ...)` so a completion can never write to a conversation other than the one that produced it. Add a regression test: start streaming in A, switch to B, let the stream complete → A unchanged, B unchanged, no duplicate created.

### S2 (HIGH): turns are only persisted on successful completion

`onSaveConversation` is called solely in `onDone` (`ChatPage.tsx:210-211`). `handleCancel` (`:415-424`) and the error path (`:220-232`) never save; nothing saves at send time; unmount cleanup (`:137-152`) aborts without saving. On minutes-long generations: Stop, an error, or closing the laptop silently discards the user's question and any partial answer (for a first turn, the entire conversation was never created). **Fix:** persist the user message at send time; persist finalized partial content in cancel and error paths (marking the message state appropriately).

### S3 (MED): unmount / engine-switch leaves `isStreaming: true`, which gets persisted

Unmount cleanup (`:137-152`) and the engine-switch effect (`:111-125`) abort without finalizing message flags (unlike `handleCancel:419-423`). The cursor renders purely off `message.isStreaming` (`ChatMessageBubble.tsx:220`); a later successful turn's save persists the stale flag → **permanent blinking cursor after reload**. **Fix:** finalize `isStreaming:false` in those paths, and additionally strip `isStreaming` in `saveMessages` before writing (it is transient UI state and should never be persisted).

### S4 (MED): `cancel()` discards buffered tokens

`lib/streaming/TokenStreamManager.ts:226-236`: `cancel()` clears the flush timer then `this.tokenBuffer = []` — never calls `flushBuffer()`, unlike `complete()` (`:144`) and `error()` (`:162`). Stop drops up to a RAF frame of tokens — potentially far more after a hidden tab (RAF is suspended; the buffer accrues). **Fix:** flush to the callback before setting cancelled. Related: on `MAX_BUFFER_SIZE` overflow, `splice(0, dropped)` (`:96-100`) silently deletes the **oldest unflushed** tokens — corrupting the middle of displayed AND persisted content; flush synchronously instead of dropping, and add a `setTimeout` fallback flush when `document.visibilityState === 'hidden'`.

### S5 (MED): `MAX_MESSAGES` prune lies, then permanently deletes

`ChatPage.tsx:362-375`: on overflow the indicator says earlier messages are "**hidden**", but the next `onDone` saves the pruned array wholesale (`updateConversation` replaces `messages`, `db/conversations.ts:57`) — the old messages are **deleted from IndexedDB**. Either keep full history in persistence and window only the render, or say "deleted" and confirm. (Also off-by-one: `[indicator, ...pruned]` yields 201 items.)

### S6 (MED): mid-stream errors are injected into markdown content

`ChatPage.tsx:221-224` appends `` `\n[Error: ${errorMessage}]` `` to `message.content` — parsed as markdown (renders inside an open code fence per A5; can even linkify). **Fix:** add an `error` field to `ChatMessage`; render a styled error card ("Something went wrong while answering. [Try again]") wired to the existing `handleRegenerate`; never mix error text into content.

### S7 (MED): WebLLM cancellation is likely inert

Stop relies on a hand-rolled `signal` field in the request type (`lib/llm/web-llm-service.ts:40-49`, used `:337-344`) that web-llm's API does not document; `interrupt()` exists on the service contract but is called **only in tests**. If the field is ignored, Stop only stops the UI while generation burns CPU/GPU to completion. **Fix:** call `llmService.interrupt()` from `cancelActiveStream`, or wire an abort listener that calls `engine.interruptGenerate()` inside `WebLLMService.generate`. (The wllama path is correct — `abortSignal` is real, `wllama-service.ts:174-224` — don't touch it.)

### S8 (LOW): relative timestamps never update

`ChatMessageList.tsx:98-101` ticks `setNow` every 60 s but discards it (`void now;` `:117` — the comment at `:113-116` asserts the opposite of what happens); `ChatMessageBubble` is `React.memo` (`:18`) with referentially-stable props on a tick, so `formatRelativeTime` (`:135,:237`) never recomputes → "Just now" forever. **Fix:** pass `now` (or a formatted label) down as a prop so it participates in memo comparison.

**Streaming facts that are already fine (do not "fix"):** token ordering is strictly preserved; `complete()` flushes synchronously before `onDone` in the same call stack — the saved array is byte-exact; StrictMode is handled deliberately (`ChatPage.tsx:167-178`, `useConversations.ts:55,120-139`); auto-scroll heuristics + Jump-to-latest are well designed and regression-tested (`ChatMessageList.tsx:36-107,263-267`); message keys are stable ids; #35's C1 (new-chat race) was **already fixed** — the non-empty→empty guard covers the New Chat path; do not regress it while fixing S1.

---

## Part 3 — UX / accessibility

Ranked for THIS deployment: non-technical users, air-gapped machines, multi-minute CPU operations.

### U1 (HIGH): the cold model load on first send is invisible

Real progress **is** wired: `ChatPage.tsx:264-268` routes `llmService.initialize()` progress into `setModelLoadingProgress`. But the only consumer is `ModelBlockedOverlay` (`:576`), which renders solely when `isModelBlocked = isBrowserMode && !isModelReady` (`:84,:572`). Readiness = file presence, not loaded-in-memory — so with packaged weights the overlay never mounts and a multi-minute ~1.3 GB load shows only the indeterminate "Generating▋" (`StreamingIndicator.tsx:66-74`). **Fix:** when a load is in flight during `isLoading`, render an inline determinate bar in the StreamingIndicator slot: "Loading the AI model — one-time, may take a few minutes… 43%" (wllama already supplies label text, `wllama-service.ts:144-152`).

### U2 (HIGH): document indexing progress is fake and uncancelable

`DocumentsPage.tsx` hardcodes milestones 30/60/75/78/90/100 (`:163,:172,:191,:206,:265,:274`). The long pole — embedding every chunk — happens between 78 and 90 in `await embeddingService.encodeBatch(texts)` (`:235`) **without** the `onProgress` callback the service already supports (`embedding-service.ts:225-228`, reports per 8-item batch). The bar freezes at 78% for minutes; users assume a hang and reload, which triggers the interrupted-processing redo. **Fix:** pass `onProgress` mapping to the 78→95 range; add stage labels ("Extracting text… / Indexing chunk 120 of 480"); add per-document cancel that aborts before the next batch.

### U3 (MED-HIGH): silent failure channels — wire the toast system that already exists

- Boot: `useServiceInitialization.ts:165-173` calls `setInitError(message)` **and** `setIsInitialized(true)` in one synchronous block; React 18 batches them into one commit; the only `initError` consumer is inside the `!isInitialized`-gated overlay (`App.tsx:60-76,120-124`) → **the error can never paint**. Search/vector init failure silently degrades retrieval. Post-boot embedding errors funnel into the same dead state (`:204-209`). **Fix:** persistent dismissible banner ("Search is degraded — answers may miss information. [Retry]") in `AppLayout`, not the boot overlay.
- Documents page: load/save/delete/quota failures are `console.error`-only (`DocumentsPage.tsx:91-93,:123-125,:144-147,:432-433`). **Fix:** error toasts.
- `ToastProvider` is mounted, tested, AA-compliant, pause-on-hover, `role="alert"` (`ToastProvider.tsx:103-145`) — and has **zero production call sites**. Wire it into: delete/rename, copy, cache-cleared, document-indexed, unsupported-file-rejected (see U7).

### U4 (MED): first-run journey ends in a guaranteed dead end

Empty-chat hero unconditionally suggests "Summarize my documents" (`ChatMessageList.tsx:16-20,:222-244`); with zero docs this runs a full multi-minute cold model load (`ChatPage.tsx:264-268` executes before the pipeline abstains) and then abstains. **Fix:** document-count-aware empty state — zero docs: "Add documents to get started" + a button navigating to Documents; docs present: grounded example prompts. The unused `EmptyState` component with action-button support already exists (`EmptyState.tsx:136-225`).

### U5 (MED): document deletion is the only destructive action without a confirm

`DocumentList.tsx:70-74,:231` fires `onDelete` in one click → `DocumentsPage.tsx:387-437` deletes doc + both indexes immediately. Misclick cost on this hardware = minutes of re-indexing. Conversation delete (`SidebarConversationItem.tsx:83-98,249-280`) and Clear Chat (`ChatPage.tsx:443-465`) both have confirm patterns — reuse one.

### U6 (MED): accessibility floor

- Light-theme status colors fail WCAG as text on `#f0f0f5`: `--color-warning #eab308` ≈ **1.7:1**, `--color-success #22c55e` ≈ **2.0:1** (`tokens.css:22-23`) — used at 11 px for exactly the labels that matter on slow hardware ("Processing...", "Ready", "Not cached", "Server not connected": `DocumentList.tsx:36-39,173-182`, `SettingsMetrics.tsx:44-52`, `ChatPage.tsx:516-527`). AA-verified `-strong` variants already exist (`tokens.css:28-29`) — apply the same pattern.
- Sidebar: focusable row sets `outline: 'none'` with no replacement (`SidebarConversationItem.tsx:176`); kebab button is `opacity: hovered ? 1 : 0` (`:218`) — an invisible keyboard tab stop (the correct `:focus-within` reveal pattern exists at `theme.css:49-60`); `role="menu"` popup (`:239`) lacks arrow-key navigation.
- Streaming live region: the entire scroll container is `role="log" aria-live="polite"` (`ChatMessageList.tsx:215`) with token-level mutations — screen readers will spam or go silent. Drop `aria-live` from the container; announce at sentence/paragraph granularity via the existing hidden announcer (`:219-221`), keep "Response complete."
- Reduced motion: no `prefers-reduced-motion` guard for `spin`/`blink`/`shimmer` (`theme.css:3-5`, `tokens.css:96-99`, `LoadingSkeleton.tsx:19-46`); boot overlay lacks `role="status"` (`App.tsx:26-78`). One CSS block + one attribute.

### U7 (MED): input/interaction fixes

- DropZone silently discards unsupported files (`DropZone.tsx:82-85` filters with no rejection callback) → surface skipped filenames (toast or the existing transient-notice pattern `DocumentsPage.tsx:548-565`).
- Header mode toggle is an air-gap footgun: one click flips to API mode (`InferenceModeToggle.tsx:28-40`), persisted across reloads (`InferenceModeContext.tsx:96-97`), jargon labels, 8 px status dot with tooltip-only meaning. Hide unless `serverUrl !== ''`; rename to plain language ("On this computer" / "Company server").
- Duplicate upload guard (`fileName+fileSize`, `DocumentsPage.tsx:327-333`) blocks legitimate re-upload with no replace path; the skip notice (`:348-354`) doesn't mention the delete-first workaround. Offer "Replace" in the notice. *(PR-B also touches this guard for the re-index flow — coordinate; the UI affordance belongs here.)*
- Keyboard shortcuts undiscoverable; **Ctrl+L hijacks the browser address bar** and double-press clears the chat (`useKeyboardShortcuts.ts:25-43`, `ChatPage.tsx:443-465,557`). Remap or drop Ctrl+L; add shortcut hints + a Settings list; document Enter vs Shift+Enter in the input placeholder (`ChatInput.tsx:272`).
- Copy button copies raw markdown source (`ChatMessageBubble.tsx:32`) — acceptable default, but offer rendered-plain-text copy alongside (derive from parsed blocks / react-markdown AST).
- Regenerate is session-ephemeral (`lastTurnRef`, `ChatPage.tsx:71,589`) — never available after reload/switch; log as accepted limitation or persist the last turn's inputs.

### U8 (MED): multimodal UI guards *(seam with PR-B — UI side lands here)*

- **Abstention-with-image:** the zero-chunk abstention (`rag-orchestrator.ts:381-398`) returns before `options.images` is ever read (`:411`) — "what's in this screenshot?" on an empty corpus abstains and the image never reaches the VLM. Guard: `contextChunks.length === 0 && !(options.images?.length)`, plus an image-only prompt branch in `buildMessages`. *(This is an orchestrator touch; keep it a minimal, isolated diff — PR-B owns the rest of that file.)*
- **Thumbnail bloat:** full-size base64 dataUrls (0.7–3.3 MB each, ×3/message) persist per message into `docqa_conversations` and the whole messages array re-serializes every turn (`types/chat.ts:7-14`, `ChatPage.tsx:345-350`, `useConversations.ts:206-209`). Persist ≤256 px JPEG thumbnails (keep full-res only in `lastTurnRef` for regenerate).
- `canAttachImages` (`ChatPage.tsx:87`) never consults `WllamaService.supportsImages()` (dead in production) — wire it or delete the method.

---

## 4 — Acceptance criteria

- [ ] Blank-line-separated ordered list renders as a single `<ol>` numbered 1..N; list starting at "3." renders 3..N (`start` attr); `"1969. It was…"` renders as a paragraph with the year intact. Tests assert semantic DOM, incl. the blank-line case (current tests only cover contiguous).
- [ ] GFM tables, `---`, nested lists, `__bold__`, `~~strike~~`, task lists render correctly (Option B) or are explicitly descoped with tests documenting chosen behavior (Option A).
- [ ] Streaming a message with an open code fence never permanently swallows subsequent content.
- [ ] Start streaming in conversation A → switch to B → completion: A unchanged, B unchanged, no duplicate conversation, no re-pointing. Regression test required.
- [ ] Stop/error/close mid-generation: user message + partial answer survive reload; no message reloads with a blinking cursor.
- [ ] Stop delivers all buffered tokens (test: push N tokens without a RAF flush, cancel, assert callback received all N).
- [ ] Boot search-init failure shows a persistent, dismissible banner with Retry; documents-page persistence errors surface as toasts.
- [ ] Cold model load on first send shows a determinate progress bar; indexing shows real per-batch progress and is cancelable.
- [ ] Zero-doc first run guides to the Documents page; no suggested prompt can lead to an abstention on interaction #1.
- [ ] Document delete requires confirmation.
- [ ] Image + empty corpus reaches the VLM instead of abstaining.
- [ ] Light-theme status text meets WCAG AA; sidebar fully keyboard-operable with visible focus; streaming does not spam screen readers.
- [ ] No regression: XSS posture (no raw HTML rendering), scheme allowlist, auto-scroll behavior, New Chat cancel guard, IME/Enter handling, wllama abort propagation.

## 5 — Seams with PR-B (coordinate, don't duplicate)

- PR-B owns `rag-orchestrator.ts` (budget/retrieval) — this PR touches it **only** for the abstention-with-image guard (U8).
- PR-B threads conversation history into the orchestrator call in `ChatPage.tsx` — the S1/S2 refactor of `runGeneration` should keep that call site clean.
- PR-B renders the `contextTrimmed` signal; this PR's error-card work (S6) shouldn't collide with that UI slot.
- The duplicate-upload guard is touched by both (U7 here, re-index flow there) — single shared helper.

## References

- Supersedes/refines #35 Clusters A & C: A1–A8 confirmed (mechanism of A1 corrected above); C1 already fixed (do not regress); C2/C3 confirmed; C4 by-design; C5 partially (virtualization gap real, "all bubbles re-render during streaming" false — memo works, don't fix a non-problem).
- Sibling: **[PR-B] Retrieval engine & air-gapped delivery**.

zaxbysauce:
Closes #36

> **Update (swarm-pr-feedback):** commit `76fab69` addresses the swarm-pr-review findings (PRR-001 BLOCKER + PRR-005 + 4 HIGH test gaps). See the closure ledger comment below.

## Summary

PR-A: chat correctness & UX overhaul for `web_ui/`. Replaces the hand-rolled markdown renderer with `react-markdown` + `remark-gfm`, fixes the CRITICAL mid-stream conversation-switch data corruption (S1) and the broader streaming/persistence defect class (S2–S8), and remediates the UX/accessibility backlog (U1–U8). Frontend-only — no vector re-index required; the only `rag-orchestrator.ts` touch is the U8a abstain-with-image guard (documented seam with #37).

**Part 1 — Renderer.** `MarkdownRenderer.tsx` rewritten on react-markdown + remark-gfm (pure-JS, air-gap-safe). Custom `urlTransform` rejects scheme-less/relative/scheme-relative URLs. No `rehype-raw` (zero-HTML-injection preserved). Code blocks get language chips + copy buttons. 100ms streaming throttle. Closes A1, N2, N3, A2, A3, A4, A5, A6, A8, N9, N10, N11, N12, A7.

**Part 2 — Streaming & persistence integrity.**
- **S1 (CRITICAL):** thread owning conversation id + messages snapshot captured at send time through `runGeneration`; `handleSend` awaits the send-time save so the owning id is resolved before streaming starts; cancel stream on conversation switch; `saveMessages` takes explicit `conversationId`. Gold-standard + first-turn regression tests prove no overwrite and no duplicate.
- **S2:** persist on send/cancel/error/unmount/engine-switch. **S3:** strip `isStreaming` before persistence; finalize flags on cleanup. **S4:** `cancel()` flushes buffered tokens before clearing; overflow flushes (no splice); hidden-tab setTimeout fallback. **S5:** render-only windowing in `ChatMessageList` (full history persists). **S6:** structured `error` field + styled error card. **S7:** WebLLM interrupt wired (abort→`interruptGenerate` listener; `cancelActiveStream` calls `interrupt()`). **S8:** thread `now` prop so timestamps recompute. Dead `appendMessage` removed.
- **PRR-001 (swarm-pr-review BLOCKER, fixed in `76fab69`):** the conversation-switch effect now PERSISTS the partial assistant turn to the owning conversation id before cancelling (previously lost — `cancel()` never fires onDone/onError). The unmount-cleanup effect is now unmount-only (`[]` dep) after discovering React runs effect destroys before creates on a dep change, which had made the prior switch-effect persist line dead code. A new `owningMessagesRef` (mirrored per-token) lets the switch/unmount/engine-switch effects read the owning conversation's partial turn instead of the live `messagesRef`.
- **PRR-005:** `handleCancel` now reads `messagesRef.current` AFTER cancel so the cancel-flushed tail tokens are persisted.

**Part 3 — UX / accessibility.** U1 determinate cold-load bar; U2 real per-batch indexing progress + cancel; U3 wired the zero-call-site toast system + persistent init-error banner; U4 document-count-aware empty state (`useDocumentCount`); U5 two-step delete confirm; U6 `--color-warning-strong` + AA labels, sidebar focus/kebab a11y, dropped `aria-live` from scroll container, `prefers-reduced-motion`; U7 DropZone feedback, mode-toggle gated on `serverUrl`, Ctrl+L→Ctrl+Shift+L; U8a image-on-empty-corpus reaches VLM, U8c `canAttachImages` consults `supportsImages()`.

## Invariant audit

- **No vector re-index:** frontend-only; single `rag-orchestrator.ts` line (U8a abstain guard). ✅
- **Zero `dangerouslySetInnerHTML` / no `rehype-raw`:** grep confirms none added. ✅
- **Scheme allowlist + tests:** `urlTransform` allows `http|https|mailto|tel`; rejects scheme-less/relative/`javascript:`/`data:`/`blob:`/`file:`. Tests assert `//evil.com/x`, `/settings`, `not-a-url` → no `<a>`. ✅
- **Auto-scroll behavior preserved:** windowing is render-only; full-array last-message + scroll heuristics intact. ✅
- **New-Chat non-empty→empty guard preserved:** `ChatPage.tsx` cancel-on-switch effect is additive. ✅
- **IME/Enter handling untouched:** only placeholder text changed. ✅
- **wllama abort path untouched:** only WebLLM `generate()` gained the abort listener. ✅
- **Air-gap packaging:** react-markdown + remark-gfm are pure-JS ESM, no runtime fetch; `validate-build.mjs` greps no CDN hostnames; `offline-env.ts` `allowRemoteModels=false` governs. ✅

## Test plan

All run from `web_ui/`:
- `npm run typecheck:all` → **0 errors** (both tsconfigs).
- `npm test` → **1117 passed, 2 skipped** (63 files). Includes the S1 switch-mid-stream + first-turn regression tests, the un-excluded `TokenStreamManager.test.ts` + `web-llm-service.test.ts` (S4/S7 contracts), and the swarm-pr-feedback additions: persist-on-switch/cancel/unmount/engine-switch, indexing cancel + progress, toast assertions, and the image-on-empty-corpus abstain guard.
- `npm run build` → **success** (only the pre-existing chunk-size warning from ML libs/weights).
- Each new regression test was verified to FAIL when its fix is reverted.

## Notes / deferred

- **U7c** (duplicate-upload "Replace" flow) deferred to PR-B (#37) — it owns the re-index flow; adding a shared helper here would expand the seam surface.
- **U7f** (regenerate persistence) logged as an accepted limitation — `lastTurnRef` is session-ephemeral by design; persisting last-turn inputs would add schema complexity for marginal benefit.
- **WASM-blocked test un-exclusions** (`DocumentsPage.test.tsx`, `ChatPage.verification-3.3.test.tsx`, etc.) remain excluded — owned by #22/#37. New tests for this PR's work live in standalone non-excluded files.
- **Cross-model review note:** the `swarm-reviewer` (Kimi K2.7) agent type was unavailable this session (requires a session restart to load its custom agent definition). The independent implementation reviewer, final critic, and PR reviewers all ran on GLM-5.2 with this limitation disclosed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ The application uses GGUF models via llama-cpp-python for fully offline inferenc
 ### Chat UI (Phase 3)
 - **Streaming Chat Interface**: Full-featured chat page (`ChatPage.tsx`) with real-time token streaming display using RAF-batched updates via `TokenStreamManager`
 - **Role-Based Message Bubbles**: Distinct styling for user, assistant, and system messages with relative timestamps ("2m ago", "just now")
-- **Inline Markdown Renderer**: Zero-dependency markdown parser supporting bold, italic, inline code, fenced code blocks, ordered/unordered lists, and links with URL validation (rejects javascript: and data: URLs)
+- **Inline Markdown Renderer**: react-markdown + remark-gfm based renderer supporting CommonMark + GFM (tables, strikethrough, task lists, autolinks, nested emphasis), fenced code blocks with language chips and per-block copy, and a URL allowlist (allows http/https/mailto/tel; rejects javascript:, data:, and scheme-less/relative URLs)
 - **Source Citation Pills**: Expandable/collapsible source pills with filename truncation, full path reveal on click, and one-click copy-to-clipboard
 - **Inference Mode Toggle**: Status indicator (green/yellow/red) for browser-local vs API mode with server connectivity check against `/auth/status` endpoint
 - **Streaming Cursor Animation**: Blinking cursor (`@keyframes blink`) appended to assistant messages during streaming for visual feedback
@@ -225,7 +225,7 @@ The application uses GGUF models via llama-cpp-python for fully offline inferenc
 | `ChatMessageList.tsx` | `src/components/` | Centered transcript (768px max-width), rich empty state with suggested prompts (Phase 2) |
 | `ChatMessageBubble.tsx` | `src/components/` | Role-based messages: assistant full-width prose, user 75% bubbles, action-row copy (Phase 2) |
 | `ChatInput.tsx` | `src/components/` | Raised card composer with focus feedback, elevation shadow, 20px radius (Phase 2) |
-| `MarkdownRenderer.tsx` | `src/components/` | Zero-dependency inline markdown parser |
+| `MarkdownRenderer.tsx` | `src/components/` | react-markdown + remark-gfm renderer (CommonMark + GFM, URL allowlist) |
 | `SourceCitation.tsx` | `src/components/` | Expandable citation pills with copy-to-clipboard |
 | `InferenceModeToggle.tsx` | `src/components/` | Status dot (green/yellow/red) for browser-local vs API mode |
 | `StreamingIndicator.tsx` | `src/components/` | Bouncing dots animation during generation |
@@ -965,7 +965,7 @@ doc_qa_app/
 │   │   │   ├── ChatMessageBubble.tsx # Role-based message bubbles (Phase 3)
 │   │   │   ├── ChatMessageList.tsx   # Scrollable message container (Phase 3)
 │   │   │   ├── ChatInput.tsx         # Input with send/cancel (Phase 3)
-│   │   │   ├── MarkdownRenderer.tsx  # Zero-dependency markdown (Phase 3)
+│   │   │   ├── MarkdownRenderer.tsx  # react-markdown + remark-gfm (Phase 3)
 │   │   │   ├── SourceCitation.tsx    # Expandable citation pills (Phase 3)
 │   │   │   ├── InferenceModeToggle.tsx # Mode status toggle (Phase 3)
 │   │   │   ├── StreamingIndicator.tsx  # Bouncing dots animation (Phase 3)
--- a/web_ui/package-lock.json
+++ b/web_ui/package-lock.json
@@ -20,6 +20,8 @@
         "pdfjs-dist": "^4.4.168",
         "react": "^18.3.1",
         "react-dom": "^18.3.1",
+        "react-markdown": "^9.1.0",
+        "remark-gfm": "^4.0.1",
         "xlsx": "^0.18.5"
       },
       "devDependencies": {
@@ -2316,6 +2318,15 @@
         "assertion-error": "^2.0.1"
       }
     },
+    "node_modules/@types/debug": {
+      "version": "4.1.13",
+      "resolved": "https://registry.npmjs.org/@types/debug/-/debug-4.1.13.tgz",
+      "integrity": "sha512-KSVgmQmzMwPlmtljOomayoR89W4FynCAi3E8PPs7vmDVPe84hT+vGPKkJfThkmXs0x0jAaa9U8uW8bbfyS2fWw==",
+      "license": "MIT",
+      "dependencies": {
+        "@types/ms": "*"
+      }
+    },
     "node_modules/@types/deep-eql": {
       "version": "4.0.2",
       "resolved": "https://registry.npmjs.org/@types/deep-eql/-/deep-eql-4.0.2.tgz",
@@ -2327,7 +2338,39 @@
       "version": "1.0.8",
       "resolved": "https://registry.npmjs.org/@types/estree/-/estree-1.0.8.tgz",
       "integrity": "sha512-dWHzHa2WqEXI/O1E9OjrocMTKJl2mSrEolh1Iomrv6U+JuNwaHXsXx9bLu5gG7BUWFIN0skIQJQ/L1rIex4X6w==",
-      "dev": true,
+      "license": "MIT"
+    },
+    "node_modules/@types/estree-jsx": {
+      "version": "1.0.5",
+      "resolved": "https://registry.npmjs.org/@types/estree-jsx/-/estree-jsx-1.0.5.tgz",
+      "integrity": "sha512-52CcUVNFyfb1A2ALocQw/Dd1BQFNmSdkuC3BkZ6iqhdMfQz7JWOFRuJFloOzjk+6WijU56m9oKXFAXc7o3Towg==",
+      "license": "MIT",
+      "dependencies": {
+        "@types/estree": "*"
+      }
+    },
+    "node_modules/@types/hast": {
+      "version": "3.0.5",
+      "resolved": "https://registry.npmjs.org/@types/hast/-/hast-3.0.5.tgz",
+      "integrity": "sha512-rp/ezSWaD1m44dPKICGhiskI13nVr7qTloFwDa/IYkhhf5nzwP+zIQcIJh3WIFSBOy/H1PzB40jPjMDksN4F+g==",
+      "license": "MIT",
+      "dependencies": {
+        "@types/unist": "*"
+      }
+    },
+    "node_modules/@types/mdast": {
+      "version": "4.0.4",
+      "resolved": "https://registry.npmjs.org/@types/mdast/-/mdast-4.0.4.tgz",
+      "integrity": "sha512-kGaNbPh1k7AFzgpud/gMdvIm5xuECykRR+JnWKQno9TAXVa6WIVCGTPvYGekIDL4uwCZQSYbUxNBSb1aUo79oA==",
+      "license": "MIT",
+      "dependencies": {
+        "@types/unist": "*"
+      }
+    },
+    "node_modules/@types/ms": {
+      "version": "2.1.0",
+      "resolved": "https://registry.npmjs.org/@types/ms/-/ms-2.1.0.tgz",
+      "integrity": "sha512-GsCCIZDE/p3i96vtEqx+7dBUGXrc7zeSK3wwPHIaRThS+9OhWIXRqzs4d6k1SVU8g91DrNRWxWUGhp5KXQb2VA==",
       "license": "MIT"
     },
     "node_modules/@types/node": {
@@ -2343,14 +2386,12 @@
       "version": "15.7.15",
       "resolved": "https://registry.npmjs.org/@types/prop-types/-/prop-types-15.7.15.tgz",
       "integrity": "sha512-F6bEyamV9jKGAFBEmlQnesRPGOQqS2+Uwi0Em15xenOxHaf2hv6L8YCVn3rPdPJOiJfPiCnLIRyvwVaqMY3MIw==",
-      "dev": true,
       "license": "MIT"
     },
     "node_modules/@types/react": {
       "version": "18.3.29",
       "resolved": "https://registry.npmjs.org/@types/react/-/react-18.3.29.tgz",
       "integrity": "sha512-ch0qJdr2JY0r04NXSprbK6TXOgnaJ1Tz23fm5W+z0/CBah6BSBc3n96h7K9GOtwh0HrilNWHIBzE1Ko4Dcw/Wg==",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@types/prop-types": "*",
@@ -2366,6 +2407,18 @@
       "peerDependencies": {
         "@types/react": "^18.0.0"
       }
+    },
+    "node_modules/@types/unist": {
+      "version": "3.0.3",
+      "resolved": "https://registry.npmjs.org/@types/unist/-/unist-3.0.3.tgz",
+      "integrity": "sha512-ko/gIFJRv177XgZsZcBwnqJN5x/Gien8qNOn0D5bQU/zAzVf9Zt3BlcUiLqhV9y4ARk0GbT3tnUiPNgnTXzc/Q==",
+      "license": "MIT"
+    },
+    "node_modules/@ungap/structured-clone": {
+      "version": "1.3.3",
+      "resolved": "https://registry.npmjs.org/@ungap/structured-clone/-/structured-clone-1.3.3.tgz",
+      "integrity": "sha512-60YRaenCQcVjYEKOcG824+DRGGIQ3VKErcBoAEDJZz5bKIs2ZG+X/H9Nk+Q6EVkwJk5QNApxbrc5QtBSwtrXAg==",
+      "license": "ISC"
     },
     "node_modules/@vitejs/plugin-react": {
       "version": "4.7.0",
@@ -2611,6 +2664,16 @@
       "dev": true,
       "license": "MIT"
     },
+    "node_modules/bail": {
+      "version": "2.0.2",
+      "resolved": "https://registry.npmjs.org/bail/-/bail-2.0.2.tgz",
+      "integrity": "sha512-0xO6mYd7JB2YesxDKplafRpsiOzPt9V02ddPCLbY1xYGPOX24NTyN50qnUxgCPcSoYMhKpAuBTjQoRZCAkUDRw==",
+      "license": "MIT",
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/wooorm"
+      }
+    },
     "node_modules/base64-js": {
       "version": "1.5.1",
       "resolved": "https://registry.npmjs.org/base64-js/-/base64-js-1.5.1.tgz",
@@ -2736,6 +2799,16 @@
       ],
       "license": "CC-BY-4.0"
     },
+    "node_modules/ccount": {
+      "version": "2.0.1",
+      "resolved": "https://registry.npmjs.org/ccount/-/ccount-2.0.1.tgz",
+      "integrity": "sha512-eyrF0jiFpY+3drT6383f1qhkbGsLSifNAjA61IUjZjmLCWjItY6LB9ft9YhoDgwfmclB2zhu51Lc7+95b8NRAg==",
+      "license": "MIT",
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/wooorm"
+      }
+    },
     "node_modules/cfb": {
       "version": "1.2.2",
       "resolved": "https://registry.npmjs.org/cfb/-/cfb-1.2.2.tgz",
@@ -2764,6 +2837,46 @@
       },
       "engines": {
         "node": ">=18"
+      }
+    },
+    "node_modules/character-entities": {
+      "version": "2.0.2",
+      "resolved": "https://registry.npmjs.org/character-entities/-/character-entities-2.0.2.tgz",
+      "integrity": "sha512-shx7oQ0Awen/BRIdkjkvz54PnEEI/EjwXDSIZp86/KKdbafHh1Df/RYGBhn4hbe2+uKC9FnT5UCEdyPz3ai9hQ==",
+      "license": "MIT",
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/wooorm"
+      }
+    },
+    "node_modules/character-entities-html4": {
+      "version": "2.1.0",
+      "resolved": "https://registry.npmjs.org/character-entities-html4/-/character-entities-html4-2.1.0.tgz",
+      "integrity": "sha512-1v7fgQRj6hnSwFpq1Eu0ynr/CDEw0rXo2B61qXrLNdHZmPKgb7fqS1a2JwF0rISo9q77jDI8VMEHoApn8qDoZA==",
+      "license": "MIT",
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/wooorm"
+      }
+    },
+    "node_modules/character-entities-legacy": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/character-entities-legacy/-/character-entities-legacy-3.0.0.tgz",
+      "integrity": "sha512-RpPp0asT/6ufRm//AJVwpViZbGM/MkjQFxJccQRHmISF/22NBtsHqAWmL+/pmkPWoIUJdWyeVleTl1wydHATVQ==",
+      "license": "MIT",
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/wooorm"
+      }
+    },
+    "node_modules/character-reference-invalid": {
+      "version": "2.0.1",
+      "resolved": "https://registry.npmjs.org/character-reference-invalid/-/character-reference-invalid-2.0.1.tgz",
+      "integrity": "sha512-iBZ4F4wRbyORVsu0jPV7gXkOsGYjGHPmAyv+HiHG8gi5PtC9KI2j1+v8/tlibRvjoWX027ypmG/n0HtO5t7unw==",
+      "license": "MIT",
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/wooorm"
       }
     },
     "node_modules/check-error": {
@@ -2805,6 +2918,16 @@
       },
       "engines": {
         "node": ">= 0.8"
+      }
+    },
+    "node_modules/comma-separated-tokens": {
+      "version": "2.0.3",
+      "resolved": "https://registry.npmjs.org/comma-separated-tokens/-/comma-separated-tokens-2.0.3.tgz",
+      "integrity": "sha512-Fu4hJdvzeylCfQPp9SGWidpzrMs7tTrlu6Vb8XGaRGck8QSNZJJp538Wrb60Lax4fPwR64ViY468OIUTbRlGZg==",
+      "license": "MIT",
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/wooorm"
       }
     },
     "node_modules/convert-source-map": {
@@ -2864,7 +2987,6 @@
       "version": "3.2.3",
       "resolved": "https://registry.npmjs.org/csstype/-/csstype-3.2.3.tgz",
       "integrity": "sha512-z1HGKcYy2xA8AGQfwrn0PAy+PB7X/GSj3UVJW9qKyn43xWa+gl5nXmU4qqLMRzWVLFC8KusUX8T/0kCiOYpAIQ==",
-      "dev": true,
       "license": "MIT"
     },
     "node_modules/data-urls": {
@@ -2885,7 +3007,6 @@
       "version": "4.4.3",
       "resolved": "https://registry.npmjs.org/debug/-/debug-4.4.3.tgz",
       "integrity": "sha512-RGwwWnwQvkVfavKVt22FGLw+xYSdzARwm0ru6DhTVA3umU5hZc28V3kO4stgYryrTlLpuvgI9GiijltAjNbcqA==",
-      "dev": true,
       "license": "MIT",
       "dependencies": {
         "ms": "^2.1.3"
@@ -2905,6 +3026,19 @@
       "integrity": "sha512-YpgQiITW3JXGntzdUmyUR1V812Hn8T1YVXhCu+wO3OpS4eU9l4YdD3qjyiKdV6mvV29zapkMeD390UVEf2lkUg==",
       "dev": true,
       "license": "MIT"
+    },
+    "node_modules/decode-named-character-reference": {
+      "version": "1.3.0",
+      "resolved": "https://registry.npmjs.org/decode-named-character-reference/-/decode-named-character-reference-1.3.0.tgz",
+      "integrity": "sha512-GtpQYB283KrPp6nRw50q3U9/VfOutZOe103qlN7BPP6Ad27xYnOIWv4lPzo8HCAL+mMZofJ9KEy30fq6MfaK6Q==",
+      "license": "MIT",
+      "dependencies": {
+        "character-entities": "^2.0.0"
+      },
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/wooorm"
+      }
     },
     "node_modules/deep-eql": {
       "version": "5.0.2",
@@ -2964,7 +3098,6 @@
       "version": "2.0.3",
       "resolved": "https://registry.npmjs.org/dequal/-/dequal-2.0.3.tgz",
       "integrity": "sha512-0je+qPKHEMohvfRTCEo3CrPG6cAzAYgmzKyxRiYSSDkS6eGJdyVJm7WaYA5ECaAD9wLB2T4EEeymA5aFVcYXCA==",
-      "dev": true,
       "license": "MIT",
       "engines": {
         "node": ">=6"
@@ -2984,6 +3117,19 @@
       "resolved": "https://registry.npmjs.org/detect-node/-/detect-node-2.1.0.tgz",
       "integrity": "sha512-T0NIuQpnTvFDATNuHN5roPwSBG83rFsuO+MXXH9/3N1eFbn4wcPjttvjMLEPWJ0RGUYgQE7cGgS3tNxbqCGM7g==",
       "license": "MIT"
+    },
+    "node_modules/devlop": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/devlop/-/devlop-1.1.0.tgz",
+      "integrity": "sha512-RWmIqhcFf1lRYBvNmr7qTNuyCt/7/ns2jbpp1+PalgE/rDQcBT0fioSMUpJ93irlUhC5hrg4cYqe6U+0ImW0rA==",
+      "license": "MIT",
+      "dependencies": {
+        "dequal": "^2.0.0"
+      },
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/wooorm"
+      }
     },
     "node_modules/dexie": {
       "version": "4.4.4",
@@ -3179,6 +3325,16 @@
         "url": "https://github.com/sponsors/sindresorhus"
       }
     },
+    "node_modules/estree-util-is-identifier-name": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/estree-util-is-identifier-name/-/estree-util-is-identifier-name-3.0.0.tgz",
+      "integrity": "sha512-hFtqIDZTIUZ9BXLb8y4pYGyk6+wekIivNVTcmvk8NoOh+VeRn5y6cEHzbURrWbfp1fIqdVipilzj+lfaadNZmg==",
+      "license": "MIT",
+      "funding": {
+        "type": "opencollective",
+        "url": "https://opencollective.com/unified"
+      }
+    },
     "node_modules/estree-walker": {
       "version": "3.0.3",
       "resolved": "https://registry.npmjs.org/estree-walker/-/estree-walker-3.0.3.tgz",
@@ -3198,6 +3354,12 @@
       "engines": {
         "node": ">=12.0.0"
       }
+    },
+    "node_modules/extend": {
+      "version": "3.0.2",
+      "resolved": "https://registry.npmjs.org/extend/-/extend-3.0.2.tgz",
+      "integrity": "sha512-fjquC59cD7CyW6urNXK0FBufkZcoiGG80wTuPujX590cB5Ttln20E2UB4S/WARVqhXffZl2LNgS+gQdPIIim/g==",
+      "license": "MIT"
     },
     "node_modules/fdir": {
       "version": "6.5.0",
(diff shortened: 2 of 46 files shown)


zaxbysauce:
## swarm-pr-review — REQUEST_CHANGES

Full review run with 6 parallel explorer lanes → 2 independent reviewers → critic challenge. Artifacts in `.zcode/pr-review/issue-36-pr38/`. (Cross-model note: Kimi K2.7 reviewer unavailable this session; all gates ran on GLM-5.2 — disclosed.)

### 🔴 BLOCKER — PRR-001 (HIGH): mid-stream conversation switch drops the partial assistant turn

Switching conversations mid-generation **irrecoverably loses the partial answer**. The switch effect (`ChatPage.tsx:137-152`) finalizes + cancels but does NOT persist (its comment claims "the in-flight stream's own onDone/onError will persist"). But `cancel()` never fires done/error callbacks, and the RAG generator early-returns on `tokenStreamManagerRef.current !== streamManager` before reaching `complete()`. The unmount-cleanup effect does **not** rescue it — React runs the switch effect (declared line 137) before the unmount-cleanup's cleanup (declared line 241) in the same commit phase, so `cancelActiveStream` nulls the ref before the unmount-cleanup's `!== null` guard evaluates.

This directly contradicts the PR's S2 acceptance criterion ("Stop/error/close mid-generation: user message + partial answer survive reload") and loses potentially minutes of CPU generation on an ordinary UX path (clicking another conversation in the sidebar).

**Fix direction:** the switch effect should persist the finalized partial turn to the owning conversation id (`owningConversationIdRef.current`), mirroring what `handleCancel` and the engine-switch effect already do. Plus a regression test.

### 🟠 Should-fix

- **PRR-005 (MEDIUM):** `handleCancel` captures the messages snapshot *before* `cancelActiveStream`; the S4 cancel-flush then reassigns `messagesRef.current` to a new array; the finalized array is derived from the stale snapshot and overwrites the correct one — losing the cancel-flushed tail tokens (bounded to one flush interval: ≤16ms visible / ≤100ms hidden tab). Fix: snapshot *after* cancel, or read `messagesRef.current` post-cancel.
- **PRR-006 (HIGH test gap):** S2 persist-on-cancel / persist-on-unmount / persist-on-engine-switch paths have zero tests. (PRR-001 shows the switch path is currently broken — the gap is also a missing-bug detector.)
- **PRRR-008 (HIGH test gap):** U2 indexing cancel (AbortController + `handleCancelIndexing`) + the 78→95 progress mapping untested. Cancel is a load-bearing user control.
- **PRR-009 (HIGH test gap):** None of the 7 DocumentsPage toast call sites have a test asserting a specific `showToast` fires.
- **PRR-010 (HIGH test gap):** U8a image-on-empty-corpus abstain guard (`&& !(options.images?.length)`) untested — `rag-orchestrator.test.ts` was not modified.

### ✅ What's sound

- **Renderer** (react-markdown + remark-gfm): XSS posture clean (no `rehype-raw`, no `dangerouslySetInnerHTML`), `urlTransform` allowlist sound and tested (`//evil.com/x`, `/settings`, `javascript:`, `data:` all rejected), semantic DOM tests genuine, broken `__bold__` test correctly strengthened.
- **S1 no-overwrite + no-duplicate fix:** verified correct (owning-id ref + snapshot threading; the F1 first-turn await fix is sound; both regression tests are TRUE regression tests).
- **S4 (cancel flushes / overflow preserves order) + S7 (abort→interruptGenerate):** implementation + tests genuine.
- **All invariants met:** no vector re-index, scheme allowlist, auto-scroll, New-Chat guard, IME/Enter, wllama abort path untouched, air-gap packaging.
- **Docs/config clean:** deps/lock/README/validate-build/vitest/tsconfig all consistent.

### Rejected findings (transparency)
PRR-002 (transient IDB miss → duplicate): unproven, no concrete trigger. PRR-003 (save-await latency): ~12ms, imperceptible vs model load. PRR-004 (unmount-cleanup wrong-id): misread of React lifecycle — the persist branch is guarded out on switch; on true unmount the id is correct. PRR-012/015/016/017: INFO/LOW.

---
Handoff artifact at `.zcode/pr-review/issue-36-pr38/feedback-handoff.md`. Happy to proceed into `swarm-pr-feedback` to close PRR-001 + the HIGH test gaps on request.

zaxbysauce:
## 🧪 swarm-pr-review — PR #38

Full review: 6 parallel explorer lanes → 2 independent reviewers. Deterministic signals: tests (63 files, 1108 passed ✅), typecheck ✅.

---

### 🛑 CRITICAL — F-1 (PRR-001, existing finding, **still unfixed**): Conversation switch data loss

**File:** `web_ui/src/pages/ChatPage.tsx:137-152`

The conversation-switch effect finalizes the in-flight assistant message (strips `isStreaming`) and calls `cancelActiveStream()`, but does **NOT** call `onSaveConversation`. The code comment says:

> *"We do NOT persist here — the in-flight stream's own onDone/onError will persist to the OWNING conversation id"*

This assumption is **incorrect**. `TokenStreamManager.cancel()` (`TokenStreamManager.ts:248-260`) only flushes the token buffer and clears timers — it **never invokes `onDone` or `onError`**. The partial assistant message lives in React state but is **never persisted to IndexedDB**.

**Impact:** Any user who clicks a different conversation in the sidebar while a response is being generated **irrecoverably loses the partial answer**. A page reload or navigation discards it entirely.

The second commit (c8524d8) addressed a **different** finding (F1: first-turn fast-complete fix). PRR-001 remains unaddressed.

**Fix direction:** The switch effect should persist the finalized turn, mirroring what `handleCancel` already does at lines 589-596:
```
onSaveConversation(owningConversationIdRef.current, finalized, ...)
```

---

### ⚠️ MEDIUM — F-2 (PRR-005, existing finding, still present): handleCancel stale snapshot

**File:** `web_ui/src/pages/ChatPage.tsx:582-597`

`handleCancel` captures `snapshot = messagesRef.current` **before** calling `cancelActiveStream()`. When cancel flushes the token buffer via `onToken`, those flushed tokens update `messagesRef.current` — but the `finalized` array is derived from the **stale** pre-cancel snapshot, overwriting the flushed-token state. Bounded to ~one flush interval of token loss (≤16ms visible / ≤100ms hidden tab).

**Fix direction:** Derive `finalized` from `messagesRef.current` *after* `cancelActiveStream()` completes, not from the pre-cancel snapshot.

---

### ⚠️ HIGH — 4 Existing Test Gaps Still Unaddressed

These four test gaps were identified in the prior review and remain unaddressed in this PR:

1. **PRR-006**: S2 persist-on-cancel / persist-on-unmount / persist-on-engine-switch paths have **zero tests**. The new `ChatPage.streaming-persistence.test.tsx` (567 lines) covers S1 (conversation-switch owning-id) and F1 (first-turn fast-complete) only.

2. **PRR-008**: U2 indexing cancel (`AbortController`) and progress state mapping in `DocumentsPage.tsx` have no test coverage. `DocumentsPage.indexing.test.tsx` exercises indexing flow and error paths but never triggers the cancel path.

3. **PRR-009**: DocumentsPage has ~11 `showToast` call sites (lines 100, 133, 300, 368, 522, 525, 669) — no test asserts a specific toast message fires. A developer could silently break toast UX and all tests would pass.

4. **PRR-010**: The U8a image-on-empty-corpus abstain guard (`rag-orchestrator.ts:386`) allows multimodal queries on empty corpus when images are attached. `rag-orchestrator.test.ts:1043-1067` covers zero-context abstention without images but **no test supplies images** to verify the non-abstention branch.

---

### 📝 MEDIUM — O-018 Claim Inaccuracy

PR body states: *"S4/S7 test files un-excluded from vitest.config.ts"*. The actual diff removes `src/lib/streaming/TokenStreamManager.test.ts` and `src/lib/llm/web-llm-service.test.ts` from the exclude list — neither of which is "S4/S7-named". The actual S4/S7 tests live in `ChatPage.streaming-persistence.test.tsx` (new file, never excluded).

### 📝 MEDIUM — S1 Test Timing Gap

The S1 conversation-switch test deliberately engineers timing where `onDone` fires before `cancelActiveStream` — the **opposite** of production ordering. The test validates the S1 owning-id fix but does **not** test the actual cancel-on-switch persistence path (which is the F-1 BLOCKER above).

### 📝 LOW — Dead Export

`db/conversations.ts:122` exports `updateTitle` — grep finds zero in-repo consumers. Consider removing if truly dead code.

---

### ✅ What The Review Confirmed Is Sound

| Area | Status |
|------|--------|
| **S1 F1 first-turn fix** | ✅ Verified correct — `handleSend` now awaits `saveMessages`, `owningConversationIdRef` set before `runGeneration` |
| **Engine-switch persist** | ✅ Confirmed working — `ChatPage.tsx:202-211` calls `onSaveConversation` |
| **Renderer XSS posture** | ✅ No `dangerouslySetInnerHTML`, no `rehype-raw`, no `innerHTML` anywhere |
| **URL allowlist** | ✅ Rejects `javascript:`, `data:`, `blob:`, `file:`, scheme-less, and relative URLs |
| **Error rendering** | ✅ Rendered as React children (escaped) — no HTML injection |
| **Tests** | ✅ 63 files, 1108 passed, 2 skipped, 0 failed |
| **Typecheck** | ✅ Clean pass |
| **Dependency air-gap** | ✅ 40+ new transitive deps (unified/micromark ecosystem); all MIT, pure ESM, no native modules |
| **Schema compatibility** | ✅ New `ChatMessage` fields optional; `LLMService` interface additive (optional method) |
| **S7 WebLLM abort** | ✅ `AbortController` → `interruptGenerate` listener correctly wired |
| **Conversation switch test** | ✅ Validates the S1 owning-id fix (no duplicate, correct target) |

---

### Summary of Required Fixes

1. **🛑 CRITICAL**: Add `onSaveConversation` call in conversation-switch effect (`ChatPage.tsx:137-152`)
2. **⚠️ MEDIUM**: Fix `handleCancel` to use `messagesRef.current` after cancel, not stale snapshot
3. **⚠️ HIGH**: Add S2 persist-path tests (cancel/unmount/engine-switch)
4. **⚠️ HIGH**: Add U2 cancel + progress tests
5. **⚠️ HIGH**: Add DocumentsPage toast assertion tests
6. **⚠️ HIGH**: Add U8a image-on-empty-corpus guard test
7. **📝 MEDIUM**: Correct PR body O-018 claim about un-excluded test files
8. **📝 LOW**: Consider removing dead `updateTitle` export

zaxbysauce:
## swarm-pr-feedback — closure ledger (commit `76fab69`)

Addressed all findings from the two swarm-pr-review comments. Each fix was verified by an independent reviewer (GLM-5.2) AND a final critic (GLM-5.2), both of which empirically confirmed the fixes and that each new test is a TRUE regression test (fails on revert). Validation: typecheck:all 0 errors, **1117 tests pass** (was 1108; +9 regression tests), build succeeds.

| id | item | outcome | fix-ref | evidence |
|----|------|---------|---------|----------|
| **PRR-001** (BLOCKER) | conversation-switch drops partial turn | **FIXED** | `76fab69` ChatPage.tsx:146-174 | switch effect now persists to owningConversationIdRef + owningMessagesRef; unmount-cleanup dep → `[]` (the prior persist line was dead code — React runs effect destroys before creates on a dep change, traced + instrumented); regression test asserts A is saved + B untouched |
| **PRR-005** (MEDIUM) | handleCancel stale snapshot loses cancel-flushed tail | **FIXED** | `76fab69` ChatPage.tsx:637-655 | reads messagesRef.current AFTER cancelActiveStream; fake-timer test asserts T3-TAIL token survives |
| **PRR-006** (HIGH gap) | persist-on-cancel/unmount/engine-switch untested | **FIXED** | `76fab69` ChatPage.streaming-persistence.test.tsx | 3 regression tests (cancel, unmount, engine-switch) — each fails when the persist line is removed |
| **PRR-008** (HIGH gap) | U2 indexing cancel + 78→95 progress untested | **FIXED** | `76fab69` DocumentsPage.indexing.test.tsx | cancel-mid-indexing test (asserts 'Indexing cancelled' terminal state) + progress-band test (asserts aria-valuenow=87 for 5/10) |
| **PRR-009** (HIGH gap) | DocumentsPage toast call sites untested | **FIXED** | `76fab69` DocumentsPage.indexing.test.tsx | showToast spy; asserts delete-success ('Document deleted','success') + addBatch-failure ('Failed to index document for semantic search.','error') |
| **PRR-010** (HIGH gap) | U8a image-on-empty-corpus abstain guard untested | **FIXED** | `76fab69` rag-orchestrator.test.ts | images + zero-chunks does NOT abstain; generateComplete called with image part; fails when `&& !(options.images?.length)` is reverted |
| **O-018** (MEDIUM, docs) | PR body "S4/S7 test files un-excluded" wording | **FIXED** | PR body edit | corrected to name TokenStreamManager.test.ts + web-llm-service.test.ts |
| **dead updateTitle** (LOW) | zero-caller export | **FIXED** | `76fab69` conversations.ts | removed; grep confirms zero callers, no barrel export |
| **PRR-002** | transient IDB miss → duplicate | **INVALID** | — | no concrete trigger; Dexie returns existing keys reliably |
| **PRR-003** | save-await latency | **INVALID** | — | ~12ms (4 IDB ops), imperceptible vs model load |
| **PRR-004** | unmount-cleanup wrong-id | **INVALID** | — | misread of React lifecycle (the critic later confirmed the deeper ordering issue that motivated the `[]` fix) |

### Reviewer/critic verdicts
- **Independent reviewer (GLM-5.2):** APPROVE — empirically traced React ordering, verified all 9 tests fail-on-revert, no regressions.
- **Final critic (GLM-5.2, fresh context):** APPROVE — independently verified React ordering vs react.dev; confirmed PRR-001 genuinely fixed (owningMessagesRef reliably non-null at switch; persist runs before cancel flush; the S1-fallback-persists-B-to-A failure mode does not materialize). Two non-blocking caveats noted below.

### Non-blocking caveats (final critic)
- **Pre-existing transient UI glitch:** on a mid-stream switch, `setMessages(A's finalized)` briefly paints A's messages into the B view until the next render corrects it. Not a persistence bug (B never receives A's data — the test proves it) and the line predates this fix, but the lifted-state overwrite is mildly untidy.
- **StrictMode double-invoke (dev-only):** the switch effect is a dep-change re-run (not double-invoked); the unmount-cleanup's guard makes the initial-mount double-invoke safe. Untested under StrictMode but guarded.

Cross-model note repeated: `swarm-reviewer` (Kimi K2.7) was unavailable this session; all gates ran on GLM-5.2 with the limitation disclosed.

PR #42, opened 2026-07-18
fix(web_ui): resurrect cross-encoder reranker, over-fetch, abstention floor, packaging integrity (Issue #37 PR-1)

### LINKED ISSUE #37: [PR-B] Retrieval engine & air-gapped delivery: resurrect the reranker, fix relevance gating/chunking/keyword search, harden offline packaging (web_ui)

zaxbysauce:
Refs #37 (PR-1 of 2 — DO NOT auto-close; PR-2 will carry the final `Closes #37`) — reranker resurrection + packaging core)

Refs #37. PR-2 (R4/R5/R6/R7/R8/P1/P2/P3/P5/§5-eval) will follow on a separate branch after this merges; the two were split per review guidance to bound per-PR risk on the ~25-file issue #37 scope. This PR is independently shippable and delivers the headline defect: **the cross-encoder reranker never ran**, and now does.

## Summary

Issue #37's first-order driver of "poor retrieval quality" was a 3-layer reranker defect (R1): the service was never initialized (R1a), its scoring math was broken even if wired (R1b — `text-classification` pipeline softmaxes a single logit to a constant 1.0 for every pair, and `[query, text]` tuples were never pair-encoded), and the model weights were not packaged (R1c). Every downstream quality knob (relevance floor R3, over-fetch promotion R2, presets' `rerank: true`) was predicated on a working reranker that did not exist, and the test suite masked it by mocking `isReady: () => true`.

This PR lands R1 atomically (per the issue's hard requirement — wiring init alone would make things *worse* by overwriting real RRF scores with constant 1.0), plus R2 (over-fetch), R3 (abstention floor), P4 (validate-build integrity), and P6 (missing-model UX). R9 (model swaps) is deferred per the issue's own OPTIONAL/gated-on-eval-harness labeling.

### R1 — reranker resurrection (atomic: a+b+c)
- **R1a:** fire-and-forget `getRerankerService().initialize()` at boot in `useServiceInitialization.ts` (after vector+keyword init; not awaited so boot latency is independent of the ~23MB reranker load). The orchestrator's existing `isReady()` gate falls back to fused results gracefully.
- **R1b:** bypass the `text-classification` `pipeline()` entirely. Load `AutoTokenizer` + `AutoModelForSequenceClassification` directly and call the model so each `[query, passage]` pair is properly pair-encoded via `text_pair` (true `[CLS] q [SEP] p [SEP]` with alternating `token_type_ids`) and scored with `sigmoid` on the single relevance logit — the scoring function the ms-marco-MiniLM-L-6-v2 model was trained for. Batched (12/batch) to bound WASM memory; `RERANK_INPUT_CAP=50` bounds per-query latency on the target i5.
- **R1c:** package the q8 ONNX. Manifest flipped `group:'optional'→'core'` and all files `required:false→true`; `prepare-models` now `fail()`s (not `warn()`s) on missing reranker. **Critical:** transformers.js v3.x maps `dtype:'q8'` → filename `model_quantized.onnx` (the `DATA_TYPES.q8 → '_quantized'` suffix), so the manifest/prepare-models/docs all reference that exact name — a non-quantized `model.onnx` would be silently ignored and the loader 404s.
- Removed the dead `deviceMemory < 8` branch from `canRerank` (Chrome caps the API at 8, so it never tripped).

### R2 — over-fetch before rerank
- New `candidateMultiplier` on `RAGQueryOptions` + presets (`fast:1`, `balanced:3`, `quality:4`). Both retrieval legs now fetch `topK * candidateMultiplier` so the cross-encoder can promote a chunk ranked below topK in one leg.
- Raised edgevec `ef_search` 50→128 (≥2× quality's fetch k). Deleted the dead `efSearch` field from `VectorSearchOptions` (edgevec 0.6's `search(query,k)` takes no per-call ef).

### R3 — abstention floor (reachable now that R1 lands)
- `MIN_CROSS_SCORE` 0.1→0.2 (sigmoid scale; relevant ms-marco pairs >0.5).
- New `DEGRADED_VECTOR_COSINE_FLOOR=0.4` backstop: when the reranker is OFF (fast preset, or model absent), drop vector hits below the cosine floor *before* fusion so out-of-corpus questions don't fill the context with noise. edgevec score polarity verified as similarity (higher=better).
- Zero-chunk abstain path + image-attached guard (from PR #38) preserved.

### P4 — validate-build integrity
- SHA-256 checksum sidecar (`manifest.checksums.json`) written by `prepare-models`, verified by `validate-build`. **Streaming** (`createReadStream` + `for await`) — the wllama GGUF is ~2.9 GB, above Node's Buffer cap, so `readFileSync` would crash.
- LFS-pointer detection on dist contents; Vite `dist/assets/` chunk existence check; louder `--no-llm` STDERR warning.

### P6 — missing-model UX (air-gap honesty)
- wllama engine + missing packaged GGUF is now a readiness **FAILURE** (was a soft recommendation). Air-gapped builds cannot download the GGUF; the previous "Preparing the model…" overlay alongside "ensure stable internet" advice was actively wrong.
- `ModelBlockedOverlay` admin-oriented headline now triggers on the failure path; `SettingsPage` "Weights are bundled" copy gated on actual presence.

### CI / packaging
- New `--no-reranker` flag (mirrors `--no-llm`) so CI builds the typecheck/test artifact without the operator-acquired q8 ONNX; writes `VITE_EXCLUDE_MODEL_GROUPS=reranker`. Production packaging (weights staged, no flag) correctly fails if the q8 ONNX is missing.
- `PackagedModelGroup` type extended with `'reranker'`.
- PACKAGING.md + public/models/README.md document the q8 staging path, optimum CLI command, and the `--no-reranker` flag.

## Invariant audit (§3 do-not-break list — all verified intact)
- **BGE query prefix on queries only** (`rag-orchestrator.ts:252`) — untouched.
- **L2-normalize at encode + cosine metric** — untouched (embedding-service not in this diff).
- **CLS pooling end-to-end** — untouched.
- **RRF rank-only fusion, k=60** (`rag-orchestrator.ts:336`) — untouched.
- **Zero-chunk abstain path + image guard** (`rag-orchestrator.ts:439`) — preserved; now reachable via R3.
- **Boot-time index persistence ordering** — untouched (vector-index init order unchanged; reranker init is fire-and-forget after keyword init).

## Test plan
All run on `fix/web-ui-retrieval-engine-airgap-issue-37` @ HEAD after the final critic fixes.

```
$ npx tsc --noEmit                        # EXIT 0
$ npx tsc --noEmit -p tsconfig.test.json  # EXIT 0
$ npx vitest run
  Test Files  63 passed (63)
  Tests       1123 passed | 2 skipped (1125)   [2 skips are pre-existing]
$ node --check scripts/validate-build.mjs  # EXIT 0
$ node --check scripts/prepare-models.mjs  # EXIT 0
$ npm run prepare-models -- --no-reranker  # EXIT 0; writes VITE_EXCLUDE_MODEL_GROUPS=reranker
$ npm run prepare-models                   # correctly FAILS ("REQUIRED assets are missing") — production safety
```

New/changed tests covering the acceptance criteria:
- **#1 isReady:** `reranker.test.ts > initialize > loads tokenizer + model directly` + `isReady returns true after successful initialization`.
- **#2 sigmoid + pair-encoding:** `pair-encodes via text_pair (true [CLS] q [SEP] p [SEP])` asserts the tokenizer was called with `text_pair: [...passages]` + query repeated per pair; `scores are sigmoid values in (0,1), non-constant` asserts scores differ (the old softmax-collapse bug made them all 1.0).
- **#3 over-fetch promotion:** `rag-orchestrator.test.ts > reranker promotes a chunk that ranked below topK in both legs` — fused list has the cross-encoder's favorite at the bottom; after rerank it becomes `chunks[0]`.
- **#4 abstention:** `degraded no-rerank path drops vector hits below the cosine floor` + existing F3 zero-chunk test.
- **R2 fetch width:** `candidateMultiplier widens the per-leg fetch` + `default candidateMultiplier is 1`.
- **P6:** `wllama engine not-cached is a readiness FAILURE with admin message`.

## What this PR does NOT do (deferred to PR-2 or per issue gating)
- R4 (keyword `tokenize:'forward'` + stopword filter), R5 (chunking), R6 (conversation history), R7 (budget residual + contextTrimmed UI), R8 (q8 embedder + Web Worker), P1 (CI smoke), P2 (airgap flag), P3 (COOP/COEP banner), P5 (storage hygiene), §5 (eval harness). All verified as real defects at HEAD; will land in PR-2.
- R9 (model swaps: arctic-embed, ettin-reranker, edgevec 0.9) — explicitly OPTIONAL/gated on the §5 eval harness in the issue; deferred per issue text.

## Review trail
- Plan critic (GLM-5.2): NEEDS_REVISION → all BLOCKERs/MAJORs addressed.
- Implementation reviewer (Kimi K2.7): BLOCKED → fixed (q8 filename mapping, streaming SHA-256).
- Final critic (GLM-5.2): NEEDS_REVISION → fixed (`--no-reranker` CI opt-out, PACKAGING/README docs, template-literal interpolation, import hoisting). Re-validated green.

### DIFF
--- a/.github/workflows/web-ui.yml
+++ b/.github/workflows/web-ui.yml
@@ -60,8 +60,15 @@ jobs:
       # Runs AFTER build/test so a missing-weights failure here doesn't block
       # the typecheck/build/test gates. The packaging validation below re-runs
       # build with the staged models so dist/models/ is populated.
+      #
+      # Issue #37 PR-1: --no-reranker because the cross-encoder q8 ONNX is an
+      # operator-acquired weight (not in this repo / not under LFS). The
+      # reranker is REQUIRED for production packaging (prepare-models without
+      # the flag fails if the q8 ONNX is absent), but CI builds the
+      # typecheck/test artifact without it. The orchestrator degrades to fused
+      # results at runtime when the reranker group is excluded.
       - name: Prepare models (real weights via LFS)
-        run: npm run prepare-models
+        run: npm run prepare-models -- --no-reranker
 
       # Rebuild so dist/models/ contains the staged model assets, then validate.
       # (The earlier Build step ran before prepare-models, so its dist/ has no
@@ -69,25 +76,36 @@ jobs:
       - name: Rebuild with staged models
         run: npm run build
 
-      # Packaging validation with --no-llm: the browser-LLM runtime + LFM2.5-VL
-      # weights are operator-acquired (multi-GB, not in this repo), so CI
-      # enforces the embeddings + ORT + wllama-WASM core only. Full validation
-      # requires staged weights (see PACKAGING.md).
-      - name: Validate build (--no-llm)
-        run: node scripts/validate-build.mjs --no-llm
+      # Packaging validation with --no-llm --no-reranker: the browser-LLM
+      # runtime + Gemma 4 E2B-it weights AND the cross-encoder q8 ONNX are
+      # operator-acquired (multi-GB / not in this repo), so CI enforces the
+      # embeddings + ORT + wllama-WASM core only. Full validation requires
+      # staged weights (see PACKAGING.md). Issue #37 PR-1 added --no-reranker
+      # (the reranker is REQUIRED for production packaging but excluded in CI).
+      - name: Validate build (--no-llm --no-reranker)
+        run: node scripts/validate-build.mjs --no-llm --no-reranker
 
-      # Regression guard for the --no-llm runtime exclusion (feedback F1):
-      # verify prepare-models --no-llm writes the env marker that makes the
-      # runtime readiness gate treat the llm group as excluded. The deeper
-      # end-to-end wiring (Vite inlining VITE_EXCLUDE_MODEL_GROUPS into the
-      # bundle, checkPackagedModels consuming it) is covered by the
-      # excluded-groups unit tests in model-manifest.test.ts (run in the Test
-      # step above), which stub the env var directly and are stable against
-      # minification (unlike a bundle grep).
-      - name: Verify prepare-models --no-llm writes the exclusion marker
+      # Regression guard for the --no-llm / --no-reranker runtime exclusions
+      # (feedback F1; Issue #37 PR-1 extended to --no-reranker): verify
+      # prepare-models writes the env marker(s) that make the runtime readiness
+      # gate treat the excluded group(s) as "not applicable" rather than
+      # "missing". The deeper end-to-end wiring (Vite inlining
+      # VITE_EXCLUDE_MODEL_GROUPS into the bundle, checkPackagedModels consuming
+      # it) is covered by the excluded-groups unit tests in model-manifest.test.ts
+      # (run in the Test step above), which stub the env var directly and are
+      # stable against minification (unlike a bundle grep).
+      #
+      # Both flags are passed because CI does not stage the operator-acquired
+      # reranker q8 ONNX or the multi-GB LLM weights (see prepare-models step
+      # above). The grep accepts the comma-separated form (--no-llm --no-reranker
+      # writes "llm,reranker") OR the single-group form.
+      - name: Verify prepare-models writes the exclusion markers
         run: |
-          node scripts/prepare-models.mjs --no-llm
+          node scripts/prepare-models.mjs --no-llm --no-reranker
           grep -q 'VITE_EXCLUDE_MODEL_GROUPS=llm' .env.production || { \
-            echo "prepare-models --no-llm did not write the exclusion marker to .env.production"; \
+            echo "prepare-models did not write the llm exclusion marker to .env.production"; \
+            exit 1; }
+          grep -q 'reranker' .env.production || { \
+            echo "prepare-models did not write the reranker exclusion marker to .env.production"; \
             exit 1; }
           rm -f .env.production
--- a/PACKAGING.md
+++ b/PACKAGING.md
@@ -42,10 +42,29 @@ This copies into `public/models/`:
 - `ort/ort-wasm-simd-threaded.jsep.wasm` + `ort-wasm-simd-threaded.jsep.mjs` —
   the exact ORT JSEP build + ESM loader Transformers.js v3 fetches, so it never
   reaches for jsdelivr
-- `reranker/ms-marco-MiniLM-L-6-v2/` — **optional** cross-encoder. If the source
-  model isn't present at the repo root the step is skipped (a warning, not an
-  error) and the app simply runs without reranking. To include it, place the
-  model at `models/ms-marco-MiniLM-L-6-v2/` (ONNX + tokenizer) before running.
+- `reranker/ms-marco-MiniLM-L-6-v2/` — **required** cross-encoder reranker
+  (Issue #37 made it mandatory: retrieval quality depends on neural reranking,
+  and without it out-of-corpus questions never abstain). `prepare-models`
+  **fails** if the source weights are absent. The reranker loads with
+  `dtype:'q8'`, which in transformers.js v3.x resolves to the filename
+  `onnx/model_quantized.onnx` (the `DATA_TYPES.q8 → '_quantized'` suffix) — you
+  MUST stage the q8-quantized ONNX under that exact name, NOT `model.onnx`
+  (a non-quantized `model.onnx` would be silently ignored and the loader 404s).
+  Produce it via the optimum CLI:
+
+  ```bash
+  pip install optimum[onnxruntime]
+  optimum-cli export onnx --model cross-encoder/ms-marco-MiniLM-L-6-v2 \
+    --quantize q8 models/ms-marco-MiniLM-L-6-v2/onnx/
+  # → writes models/ms-marco-MiniLM-L-6-v2/onnx/model_quantized.onnx (~23 MB)
+  ```
+
+  Copy the tokenizer/config alongside it, then place the directory at
+  `models/ms-marco-MiniLM-L-6-v2/` before running `prepare-models`.
+
+  For CI / embeddings-only / server-mode builds that deliberately omit the
+  reranker, pass `--no-reranker` (mirrors `--no-llm`); the orchestrator then
+  degrades to fused results at runtime.
 
 ## 3. Build the offline archive
 
--- a/web_ui/public/models/README.md
+++ b/web_ui/public/models/README.md
@@ -19,12 +19,13 @@ public/models/
 ├── ort/                              # ONNX Runtime WASM (env.backends.onnx.wasm.wasmPaths)
 │   ├── ort-wasm-simd-threaded.jsep.wasm   # the JSEP build transformers.js v3 actually fetches
 │   └── ort-wasm-simd-threaded.jsep.mjs    # its ESM loader (also fetched same-origin)
-├── reranker/                         # OPTIONAL cross-encoder (app degrades gracefully if absent)
-│   └── ms-marco-MiniLM-L-6-v2/
+├── reranker/                         # REQUIRED cross-encoder reranker (Issue #37)
+│   └── ms-marco-MiniLM-L-6-v2/       # prepare-models fails if absent; --no-reranker opts out for CI
 │       ├── config.json
 │       ├── tokenizer.json
 │       ├── tokenizer_config.json
-│       └── onnx/model.onnx
+│       └── onnx/model_quantized.onnx # q8 ONNX; transformers.js dtype:'q8' resolves to this
+│                                      # exact name (NOT model.onnx — see PACKAGING.md §2)
 ├── wllama/                           # wllama WASM runtime (browser LLM engine)
 │   ├── wasm/wllama.wasm              # passed to wllama as AssetsPathConfig.default
 │   └── compat/                       # offline fallback for browsers w/o JSPI/Mem64
--- a/web_ui/public/models/manifest.json
+++ b/web_ui/public/models/manifest.json
@@ -26,14 +26,14 @@
     },
     {
       "id": "ms-marco-MiniLM-L-6-v2",
-      "label": "cross-encoder/ms-marco-MiniLM-L-6-v2 (reranker, optional)",
+      "label": "cross-encoder/ms-marco-MiniLM-L-6-v2 (reranker)",
       "kind": "reranker",
-      "group": "optional",
+      "group": "core",
       "files": [
-        { "path": "reranker/ms-marco-MiniLM-L-6-v2/onnx/model.onnx", "required": false },
-        { "path": "reranker/ms-marco-MiniLM-L-6-v2/tokenizer.json", "required": false },
-        { "path": "reranker/ms-marco-MiniLM-L-6-v2/config.json", "required": false },
-        { "path": "reranker/ms-marco-MiniLM-L-6-v2/tokenizer_config.json", "required": false }
+        { "path": "reranker/ms-marco-MiniLM-L-6-v2/onnx/model_quantized.onnx", "required": true, "comment": "q8 ONNX — transformers.js dtype:'q8' resolves to model_quantized.onnx (DATA_TYPES.q8 -> '_quantized' suffix). A non-quantized model.onnx would be ignored at runtime." },
+        { "path": "reranker/ms-marco-MiniLM-L-6-v2/tokenizer.json", "required": true },
+        { "path": "reranker/ms-marco-MiniLM-L-6-v2/config.json", "required": true },
+        { "path": "reranker/ms-marco-MiniLM-L-6-v2/tokenizer_config.json", "required": true }
       ]
     },
     {
--- a/web_ui/scripts/prepare-models.mjs
+++ b/web_ui/scripts/prepare-models.mjs
@@ -19,6 +19,8 @@
  */
 
 import { existsSync, mkdirSync, copyFileSync, readdirSync, statSync, writeFileSync, readFileSync } from 'node:fs';
+import { createReadStream } from 'node:fs';
+import { createHash } from 'node:crypto';
 import { dirname, join, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { isLfsPointer } from './lib/lfs-detect.mjs';
@@ -33,6 +35,14 @@ const PUBLIC_MODELS = join(WEB_UI, 'public', 'models');
 // readiness gate (checkPackagedModels) does not flag the deliberately-absent
 // browser-LLM runtime + weights as missing.
 const NO_LLM = process.argv.slice(2).includes('--no-llm');
+// `--no-reranker`: build without the cross-encoder reranker weights. Issue #37
+// made the reranker REQUIRED for production packaging (retrieval quality depends
+// on it), but CI builds on a fresh checkout do not stage the q8 ONNX (it is an
+// operator-acquired weight, not in the repo). This flag lets CI produce a
+// typecheck/build artifact without the reranker, while production packaging
+// (which runs with weights staged) omits the flag and hard-fails if the q8
+// ONNX is missing — catching a real packaging defect.
+const NO_RERANKER = process.argv.slice(2).includes('--no-reranker');
 
 let hadError = false;
 
@@ -113,9 +123,12 @@ function prepareEmbeddingModel() {
 }
 
 // ---------------------------------------------------------------------------
-// 1b. Reranker model (OPTIONAL): cross-encoder/ms-marco-MiniLM-L-6-v2.
-//     Reranking degrades gracefully when absent, so a missing source is a warning,
-//     not a build failure.
+// 1b. Reranker model: cross-encoder/ms-marco-MiniLM-L-6-v2.
+//     Issue #37 R1c: the reranker is REQUIRED for retrieval quality. A missing
+//     source is now a HARD FAILURE (matches embedding-model semantics) so CI
+//     cannot silently ship a build with the reranker absent. Runtime fallback
+//     still applies if init fails at load time (the orchestrator's isReady()
+//     gate degrades gracefully), but packaging must not skip it.
 // ---------------------------------------------------------------------------
 function prepareRerankerModel() {
   const srcDir = firstExisting([
@@ -126,25 +139,40 @@ function prepareRerankerModel() {
   const destDir = join(PUBLIC_MODELS, 'reranker', 'ms-marco-MiniLM-L-6-v2');
 
   if (!srcDir) {
-    warn(
-      'reranker model source not found (optional). Cross-encoder reranking will be ' +
-        'disabled in the offline build. See PACKAGING.md to include it.'
+    fail(
+      'reranker model source not found. Expected models/ms-marco-MiniLM-L-6-v2/ ' +
+        'at the repo root (or models/cross-encoder/ms-marco-MiniLM-L-6-v2/). ' +
+        'Issue #37 made the reranker required — retrieval quality depends on it. ' +
+        'See PACKAGING.md for how to stage the q8 ONNX (~23MB).'
     );
     return;
   }
 
-  const onnx = join(srcDir, 'onnx', 'model.onnx');
+  // transformers.js dtype:'q8' resolves to `model_quantized.onnx` (the
+  // DATA_TYPES.q8 -> '_quantized' filename suffix). Stage the q8 ONNX under
+  // that exact name; a non-quantized model.onnx would be silently ignored at
+  // runtime and the reranker init would 404.
+  const onnx = join(srcDir, 'onnx', 'model_quantized.onnx');
   if (!existsSync(onnx) || statSync(onnx).size < 1024) {
-    warn(`reranker ONNX missing or an LFS stub at ${onnx}; skipping (optional).`);
+    fail(
+      `reranker q8 ONNX weights missing or look like an LFS stub: ${onnx} ` +
+        '(size < 1KB). The reranker loads with dtype:\'q8\', which expects ' +
+        'onnx/model_quantized.onnx (NOT model.onnx). Stage the q8 ONNX under ' +
+        'that exact name — see PACKAGING.md.'
+    );
     return;
   }
   if (isLfsPointer(onnx)) {
-    warn(`reranker ONNX is a Git-LFS pointer stub at ${onnx}; skipping (optional). Run \`git lfs pull\`.`);
+    fail(
+      `reranker q8 ONNX weights are a Git-LFS pointer stub (not the real file): ${onnx}. ` +
+        'Run `git lfs pull` (or copy the model from a machine that has the real weights) ' +
+        'before packaging. See PACKAGING.md.'
+    );
     return;
   }
 
   copyTree(srcDir, destDir);
-  log(`reranker (optional) -> ${destDir}`);
+  log(`reranker -> ${destDir}`);
 }
 
 // ---------------------------------------------------------------------------
@@ -273,7 +301,11 @@ function prepareOnnxRuntimeWasm() {
 // ---------------------------------------------------------------------------
 log(`assembling offline model assets${NO_LLM ? ' (--no-llm: embeddings-only / server mode)' : ''}...`);
 prepareEmbeddingModel();
-prepareRerankerModel();
+if (!NO_RERANKER) {
+  prepareRerankerModel();
+} else {
+  log('--no-reranker: skipping cross-encoder reranker weights (CI build). The orchestrator degrades to fused results without it; production packaging MUST omit this flag.');
+}
 if (!NO_LLM) {
   prepareWllamaRuntime();
   prepareBrowserLLM();
@@ -282,13 +314,19 @@ if (!NO_LLM) {
 }
 prepareOnnxRuntimeWasm();
 
-// When --no-llm is set, record the excluded group so the runtime readiness gate
-// (checkPackagedModels) treats the llm group as "not applicable" rather than
-// "missing". Vite loads `.env.production` for `vite build`, so writing here makes
-// the value reach `import.meta.env.VITE_EXCLUDE_MODEL_GROUPS` at build time.
+// When --no-llm / --no-reranker is set, record the excluded group(s) so the
+// runtime readiness gate (checkPackagedModels) treats the deliberately-absent
+// group(s) as "not applicable" rather than "missing". Vite loads
+// `.env.production` for `vite build`, so writing here makes the value reach
+// `import.meta.env.VITE_EXCLUDE_MODEL_GROUPS` at build time. The env value is a
+// comma-separated list (model-manifest.ts splits on ',').
 const ENV_FILE = join(WEB_UI, '.env.production');
 const MARKER = 'VITE_EXCLUDE_MODEL_GROUPS=';
-const desiredLine = NO_LLM ? `${MARKER}llm` : '';
+const excludedGroups = [
+  ...(NO_LLM ? ['llm'] : []),
+  ...(NO_RERANKER ? ['reranker'] : []),
+];
+const desiredLine = excludedGroups.length > 0 ? `${MARKER}${excludedGroups.join(',')}` : '';
 let envLines = [];
 if (existsSync(ENV_FILE)) {
   envLines = readFileSync(ENV_FILE, 'utf8').split(/\r?\n/).filter((l) => !l.startsWith(MARKER) && l.trim() !== '');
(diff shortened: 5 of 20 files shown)


zaxbysauce:
## 🧪 swarm-pr-review — PR #42

**Verdict: APPROVED** — 2 LOW findings (advisory, non-blocking).

Review: 6 parallel explorer lanes → 1 independent reviewer → 1 critic challenge. Deterministic signals: typecheck PASS ✅, CI latest run PASS ✅, 50/77 test files failing (ALL pre-existing, same count without PR changes).

---

### 📝 LOW — Advisory findings (not blocking)

**F-1: Malformed checksum sidecar silently ignored**
`web_ui/scripts/validate-build.mjs:121-131`

The try/catch around `JSON.parse` for the `manifest.checksums.json` sidecar silently falls through when the file is malformed. If both candidate sidecars fail parsing, `checksums` stays `{}`, lines 165-168 produce no `checksumJobs`, and lines 173-188 skip all SHA-256 verification. The packaging integrity guarantee becomes a no-op without warning.

**F-2: RERANK_INPUT_CAP test doesn't prove the cap**
`web_ui/src/lib/search/reranker.test.ts:476-510`

The test feeds 60 inputs with `RERANKER_BATCH_SIZE=12`. Without the 50-item cap: 60 pairs → ceil(60/12) = 5 batches. With the cap: 50 pairs → ceil(50/12) = 5 batches. The assertion `expect(pairsScored).toBe(5)` matches both scenarios and does not prove the 50-item cap is actually enforced.

---

### ✅ Verified Sound

| Category | Result |
|----------|--------|
| **R1a (boot init)** | `useServiceInitialization.ts:158` calls `getRerankerService().initialize()` fire-and-forget after vector+keyword init |
| **R1b (correct scoring)** | `reranker.ts:203-204` uses `AutoTokenizer.from_pretrained` + `AutoModelForSequenceClassification.from_pretrained` directly, `text_pair` at line 312, `sigmoid` at line 319 |
| **R1c (packaged weights)** | `manifest.json:27-37` flips reranker group to `"core"`, all files `required: true`; `prepare-models.mjs` now `fail()`s on missing reranker |
| **R2 (over-fetch)** | `rag-presets.ts:19,21,24` — fast:1, balanced:3, quality:4; `rag-orchestrator.ts:200-201` computes `fetchK = topK * candidateMultiplier` |
| **R2 (ef_search 128)** | `vector-index.ts:121` — `config.ef_search = 128`; dead `efSearch` field deleted from `types/search.ts` |
| **R3 (abstention floor)** | `rag-orchestrator.ts:132` — `MIN_CROSS_SCORE = 0.2`; line 150 — `DEGRADED_VECTOR_COSINE_FLOOR = 0.4`; applied pre-fusion at line 324 |
| **P4 (packaging integrity)** | SHA-256 streaming (`createReadStream` + `for await`), LFS pointer detection, Vite chunk check — all in `validate-build.mjs` |
| **P6 (missing-model UX)** | `model-readiness.ts` — wllama missing GGUF is a readiness FAILURE, admin-oriented message path verified |
| **CI --no-reranker** | `web-ui.yml:71,86,104` passes `--no-reranker`; `prepare-models.mjs:38-45` handles the flag; writes `VITE_EXCLUDE_MODEL_GROUPS=reranker` |
| **All 6 invariants** | BGE query prefix, L2-normalize + cosine, CLS pooling, RRF k=60, zero-chunk abstention, boot ordering — all confirmed untouched |
| **Security** | CLEAN — all paths from hardcoded constants; `configureOfflineEnv()` prevents remote fetches; no secrets, credentials, or injection vectors |
| **Failed init retry** | `reranker.ts:226` clears `initPromise` before rejecting, so subsequent `initialize()` calls properly retry |
| **didRerank flag** | Set only AFTER awaited `rerank()` succeeds (`rag-orchestrator.ts:364-381`) — no race condition |
| **Empty-text alignment** | Empty-text items excluded before tokenizer batching (`reranker.ts:287-296`), arrays stay aligned |
| **Tokenizer leak** | Transformers.js tokenizer is JS object construction with no disposal API; ONNX session owned by model, disposed correctly at `reranker.ts:217-223` |
| **Test coverage** | `reranker.test.ts` verifies `AutoTokenizer`/`AutoModel` factories; `rag-orchestrator.test.ts` verifies cosine floor pre-fusion |

### Disproved Candidates
- Lane 4 (docs-intent) candidates C-001 through C-011 — all DISPROVED. The explorer hallucinated missing implementations. ALL PR features verified present on disk.

### Suggested Fixes (non-blocking)
1. **F-1**: Treat a malformed checksum sidecar as a build failure (or require a successfully parsed alternative path).
2. **F-2**: Add a cap test that records total tokenizer passages and asserts exactly 50.

### Phase 0B — Branch State
- `mergeStateStatus: BLOCKED` (likely branch protection / check requirements — not a conflict)
- `mergeable: MERGEABLE` — no merge conflicts
- No review comments or reviews on this PR
- CI latest run: PASS
- Handoff: `swarm pr-feedback https://github.com/ZaxbyHub/trainingapp/pull/42 continue from .swarm/pr-review/pr42/feedback-handoff.md` (if fixes are desired)

Issue #72, opened 2026-09-05
[Workstream C] PR 5 of 9: Grounded/general provenance field in both APIs and a visible chat UI badge


zaxbysauce:
**Roadmap:** C5 · **Workstream C: Knowledge packs and freshness**
**Sequence:** PR 5 of 9 (tracking slot; dependencies define execution order)
**Roadmap baseline:** 2026-09-05 · **Priority:** P1 · **Repo revision:** b2ddda06e86b63f81bdbedce220a3a2357bc9852

## Outcome
When this closes, every answer from either backend is explicitly labeled as `grounding: "grounded"` (evidence cleared the relevance floor) or `grounding: "general"` (the LLM answered from general knowledge because no evidence cleared the floor), and the chat UI in both server-mode and browser-mode renders a visible badge reflecting it. This issue does NOT change the relevance floor value itself or retrain/re-tune it — it wires an existing floor decision into a new, explicit, user-visible field. It does NOT implement the recency/dedup ranking that feeds it (C4 — dependency, not scope here).

## Evidence motivating this slot
- CONFIRMED: [rag_engine.py:566-585](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/rag_engine.py#L566-L585) region (per audit_report.md 1.5) already has a canned no-chunks response and a secondary fallback (:627-646) "when the LLM claims ignorance despite sources" — these are the two behavior branches this issue must attach an explicit machine-readable `grounding` value to; today they are prose-only, not a structured field consumers can badge.
- CONFIRMED (audit_report.md 2.5): web_ui's `rag-orchestrator.ts:500-517` already implements an abstain + relevance-floor check with `MIN_CROSS_SCORE=0.2` (rag-orchestrator.ts:148-163, documented as un-revalidated for the ettin reranker) — the floor mechanism exists on the browser side too; this issue formalizes its output as the same `grounding` enum, not a new threshold.
- CONFIRMED (issues_draft.md, authoritative dependency graph): "C5 ... Depends on: C4" — grounding must reflect the POST-recency, POST-dedup evidence set, not the raw pre-fusion candidate set, otherwise a chunk that C4 would exclude as superseded could still count as "grounded" evidence.

## Prerequisites and ownership
[...]

Issue #50, opened 2026-09-05
[Workstream 0] Epic: Trainingapp v3 — Electron local-first RAG + OpMed Learning Companion


zaxbysauce:
**Roadmap:** Epic · **Trainingapp v3 — Electron local-first RAG + OpMed Learning Companion**
**Roadmap baseline:** 2026-09-05 · **Repo revision:** b2ddda0 · **Project board:** https://github.com/orgs/ZaxbyHub/projects/1

## Outcome

A Windows installer for a 12th-gen Intel i5 / 16 GB RAM / 256 GB SSD laptop (no discrete GPU) that answers tier-0 questions about the target application from bundled documentation packs, user-added packs/documents and the OpMed CDP MicroLearning Companion training, preferring newer content, at >= 8 tok/s decode and <= 3 s first token on the reference laptop, with an embedded Storyline player and a Learn panel that opens the exact training slide for the answer. Browser (wllama) remains a secondary, degraded surface; the Python backend stays conformant to the same API contract.

## Evidence base

The first four comments on this issue hold the critic-corrected audit report, the Storyline package analysis, the architecture plan and the independent GPT-5.6 Sol critic verdict. Key verified facts: `/ask/stream` is broken today (coroutine never awaited, api_server.py:722-745); the renderer's "server mode" covers chat only; the desktop LLM refuses to load unless ~12.4 GB RAM is free and reports a misdirecting error; no performance numbers exist anywhere in the repo; bundled-docs support was built then deleted; the desktop reranker is inert on shipped installs; the browser re-reads the ~3.85 GB model on every page load. Each is owned by a slot below.

## Workstream map and join points

- **A (8)** gates everything: #51 repairs and freezes the API contract; #57 decides Node backend vs Python sidecar (ADR-0003); #58 proves Storyline slide jumps; #55/#56 fix the model choices by ADR. A2/A3/A4 are parallel-safe.
- **B (9)** is the critical path once #51 and #57 close; B5 (store interop) may start after #55.
- **C (9)**: C1/C2 start immediately; C3+ need B5–B7.
- **D (7)**: D1/D2 start after C1; D3 needs C6; D4–D7 need #58 and B.
- **E (6)** closes the roadmap.

[...]

Issue #40, opened 2026-07-17
Answer quality: no conversation memory, degenerate repetition, irrelevant follow-up retrieval (web_ui)


zaxbysauce:
## Context for the implementer

`web_ui/` is the sole release artifact: one HTML5 app, fully air-gapped (zero runtime network), running in-browser on 12th-gen Intel i5 CPUs with no dedicated GPU. Stack: React 18 + Vite + TS, wllama (llama.cpp WASM) LLM, transformers.js embeddings, Dexie/IndexedDB persistence. Generations take **minutes** on target hardware.

Reported symptoms (from live testing against a medication-ordering document corpus):
1. **"Doesn't remember"** — follow-up questions get no context from prior turns. "Right but how do I fix it?" after asking about medication ordering returns a completely unrelated answer about data storage.
2. **Degenerate repetition** — the model repeats itself: "If you have a cosign for orders, ensure that the cosigner is assigned to the order. If you have a cosign for orders, ensure that the cosigner is assigned to the orders." (4× with slight variations).
3. **Irrelevant follow-up answers** — "how do I fix it?" retrieves chunks about "Data & infrastructure Storage & hosting" (unrelated), because the retriever embeds the literal pronoun-heavy string with zero context.

These are **pre-existing design gaps** (confirmed via git archaeology: history was never threaded since the first commit). The model swap to Gemma 4 E2B-it (PR #39) helps but does **not** fix them — a better model still gives stateless, repetitive, irrelevant answers until the code is fixed.

Scope: everything under `web_ui/src/`. This issue is independent of the retrieval-engine work in #37.

---

## Root cause 1 — Conversation history is NOT threaded into the LLM (CRITICAL)

The RAG orchestrator is **single-turn by design**. `RAGOrchestrator.query()` (`src/lib/rag/rag-orchestrator.ts:164`) accepts only `question: string` — no history parameter. `buildMessages()` (`:505-527`) emits exactly two messages:

```ts
return [
  { role: 'system', content: systemPrompt },
  { role: 'user', content: `Context:\n${context}\n\nQuestion: ${question}` },
];
```

`ChatPage.runGeneration` (`src/pages/ChatPage.tsx:478`) passes only the latest `text` string. The `RAGQueryOptions` interface (`:42-61`) has **no `history`/`messages`/`priorTurns` field**. Every turn is an isolated stateless Q&A.

[...]

Issue #61, opened 2026-09-05
[Workstream B] PR 3 of 9: Backend host in Electron implementing the frozen API contract, per ADR-0003


zaxbysauce:
**Roadmap:** B3 · **Workstream B: Electron shell and desktop backend**
**Sequence:** PR 3 of 9 (tracking slot; dependencies define execution order)
**Roadmap baseline:** 2026-09-05 · **Priority:** P0 · **Repo revision:** b2ddda0

## Outcome
Closing this gives an Electron-hosted backend — implemented per whichever option ADR-0003 (#57) selects — that answers the frozen contract from #51 (A1) and passes the same conformance suite `api_server.py` passes. This does NOT certify model quality, streaming performance, or memory budget (B4/#62, B8/#66); it certifies wiring and conformance only, with stub handlers acceptable for routes not yet backed by real inference.

## Evidence motivating this slot
- The contract is frozen by #51 (A1), which repairs `api_server.py:722` (`asyncio.to_thread(...)` assigned without `await`/`Task`, then `.done()`/`.result()` called at `:736,745`) per `implementation_plan.md:3` and `audit_report.md` gap #13 — CONFIRMED against the plan docs; re-verify exact lines at implementation time (this issue's scope is `desktop/`, consuming `contracts/api.openapi.yaml`, not re-diagnosing the Python bug).
- [web_ui/src/pages/ChatPage.tsx:437-453](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/web_ui/src/pages/ChatPage.tsx#L437-L453) — CONFIRMED: `if (mode === 'api') { ... url = serverUrl ? \`${serverUrl...}/ask/stream\` : '/ask/stream'; streamManager.startSSEStream(url, { question: text, history: buildHistorySnapshot(owningMessages) }, getToken() ?? undefined); }` — this is the first real conformance target: `question`+`history` via SSE at `/ask/stream`.
- [web_ui/src/lib/api/index.ts:40](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/web_ui/src/lib/api/index.ts#L40) — CONFIRMED unused today; the backend host must still answer every route `ApiClient` defines since B9 wires it later.
[...]

PR #90, opened 2026-09-05
[WS-A PR 1/8] Repair and freeze the renderer/Python API contract (fix /ask/stream, SSE lifecycle, cancel, CORS, loopback default)

### LINKED ISSUE #51: [Workstream A] PR 1 of 8: Repair /ask/stream, freeze the API contract, and lock down default bind/auth
**Roadmap:** A1 · **Workstream A: Contract, evidence and decisions**
**Sequence:** PR 1 of 8 (tracking slot; dependencies define execution order)
**Roadmap baseline:** 2026-09-05 · **Priority:** P0 · **Repo revision:** b2ddda06e86b63f81bdbedce220a3a2357bc9852

## Problem
`/ask/stream` in `api_server.py` has never worked. [api_server.py:722-728](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/api_server.py#L722-L728) assigns the return of `asyncio.to_thread(...)` — a bare coroutine — directly to `query_future`, without `await`ing it or wrapping it in `asyncio.create_task(...)`. The poll loop then calls `query_future.done()` at [api_server.py:736](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/api_server.py#L736) and `query_future.result()` at [api_server.py:745](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/api_server.py#L745), neither of which exist on a coroutine object — this raises `AttributeError` on the first idle poll (0.1 s in) and the query never runs. Every other endpoint correctly awaits `to_thread` directly (`/ask` at [api_server.py:530](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/api_server.py#L530), and two more `to_thread` call sites at lines 543 and 572 per audit_report.md). `tests/test_api_asyncio_thread.py` mocks `to_thread` and only exercises `/ask`; CONFIRMED it has zero occurrences of "stream" (grepped the file, no matches) — so this defect ships green through CI today.

Consequence: `web_ui`'s "server mode" chat POSTs to `/ask/stream` unconditionally ([ChatPage.tsx:437-453](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/web_ui/src/pages/ChatPage.tsx#L437-L453): `const url = ... '/ask/stream'; streamManager.startSSEStream(url, { question: text, history: buildHistorySnapshot(owningMessages) }, ...)`) — server-mode chat against the Python backend has never functioned end to end.

Separately, [api_server.py:1002](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/api_server.py#L1002) defaults `API_HOST` to `"0.0.0.0"` (`host = os.environ.get("API_HOST", "0.0.0.0")  # nosec: B104 — intentional`) while [auth.py:25](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/auth.py#L25) defaults `ENABLE_AUTH` false — an unauthenticated RAG endpoint exposed on the LAN if the server is started as documented.

This issue does NOT certify performance, does NOT implement the Node/Python backend decision (A7), and does NOT change renderer surfaces other than the SSE client behavior needed to consume the corrected stream lifecycle.

## Requested change
Fix the coroutine bug, harden the SSE lifecycle (cancellation, terminal events), tighten the default bind, and — because every downstream Electron/Node backend (A7, WS-B) and conformance test (A4, E4) must target one frozen shape — extract and pin the full API surface into `contracts/api.openapi.yaml` plus a backend-agnostic conformance suite.

- CONFIRMED [api_server.py:722-728](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/api_server.py#L722-L728): `query_future = asyncio.to_thread(engine.query, request.question, n_results=request.n_results, stream_callback=stream_callback, conversation_history=request.history)` — coroutine, not a Task.
- CONFIRMED [api_server.py:736](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/api_server.py#L736): `if query_future.done():` — `AttributeError: 'coroutine' object has no attribute 'done'`.
- CONFIRMED [api_server.py:745](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/api_server.py#L745): `result = query_future.result()  # Raises if query failed` — same defect class.
- CONFIRMED [api_server.py:1002](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/api_server.py#L1002): `host = os.environ.get("API_HOST", "0.0.0.0")`.
- CONFIRMED [auth.py:25](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/auth.py#L25): `ENABLE_AUTH = os.environ.get("ENABLE_AUTH", "false").lower() == "true"`.
- CONFIRMED [ChatPage.tsx:443-454](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/web_ui/src/pages/ChatPage.tsx#L443-L454): server mode always targets `/ask/stream` with body `{ question, history }`.
- CONFIRMED route inventory (grepped `@app.get`/`@app.post`/`@app.delete`/`@app.put`): `GET /` (467), `GET /auth/status` (483), `POST /auth/token` (489), `GET /stats` (514), `POST /ask` (530), `POST /search` (563), `POST /ingest` (582), `POST /ingest/file` (611), `DELETE /documents` (672), `GET /documents` (688), `POST /ask/stream` (699, gated on `HAS_SSE`), `POST /ingest/batch` (778), `GET /settings` (888), `PUT /settings` (908).
- CONFIRMED `QuestionRequest` [api_server.py:179-192](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/api_server.py#L179-L192) (`question` 1-2000 chars, `n_results` 1-10 default 6, `history: Optional[List[Dict[str,str]]]` max 20 items); `QuestionResponse` [api_server.py:235-243](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/api_server.py#L235-L243) (`question, answer, sources: List[str], context_length, inference_time`) — no `grounding`/`learn` fields exist today.
- CONFIRMED CORS [api_server.py:412-416](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/api_server.py#L412-L416): `allow_methods=["GET","POST","DELETE"]` — `PUT` (used by `/settings`) is missing, a live cross-origin gap.
- CONFIRMED `web_ui/src/lib/api/client.ts` has NO `/ask/stream` method; streaming goes through the separate `SSEStreamConsumer` in `streaming.ts`. `types.ts` has no `history`/`grounding`/`learn` fields even though Python's `QuestionRequest.history` already exists.
- CONFIRMED `streaming.ts`'s terminal-event contract: `done` = `data.sources !== undefined && data.context_length !== undefined` ([streaming.ts:369-375](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/web_ui/src/lib/api/streaming.ts#L369-L375)); token = `data.token !== undefined` (:376-378); explicit `data.error` handled (:379-381); a close with no terminal event synthesizes an error (:334-336). `contracts/` must encode this discriminated union exactly.
- INFERRED: no test exercises the corrected `/ask/stream` end-to-end; `tests/test_api_asyncio_thread.py` never calls the streaming route.

No dependencies — the gate for the whole roadmap. #57 (A7) depends on this for the frozen contract both candidate backends prove against. #54 (A4) depends on this for a stable route shape. #61 (B3) depends transitively on the OpenAPI file and conformance suite. Owns: the coroutine fix, SSE lifecycle, CORS fix, bind hardening, `contracts/api.openapi.yaml`, `contracts/tests/`. Does not own: Electron/Node implementation (B3), full renderer inventory (B9), performance measurement (A2).

## Required scope
- [ ] `api_server.py`: wrap the query in a real awaitable — `query_task = asyncio.create_task(asyncio.to_thread(engine.query, ...))` — and replace `.done()`/`.result()` calls on `query_task` accordingly.
- [ ] Propagate client disconnect: use `await request.is_disconnected()` inside the poll loop to cancel `query_task` and the `engine.query` cancellation token, so an aborted fetch actually stops CPU work.
- [ ] Emit terminal events matching `streaming.ts`'s discriminated union: token `{token}`, success `{sources, context_length, inference_time}`, cancellation/error payloads with an explicit `error` key.
- [ ] Fix `api_server.py:412-416` CORS `allow_methods` to include `PUT` (required by `/settings`).
- [ ] Change `API_HOST` default from `0.0.0.0` to `127.0.0.1` in `api_server.py:1002`; require explicit opt-in to bind non-loopback; document in `CONFIGURATION.md`.
- [ ] Add a `web_ui/src/lib/api/client.ts` `askStream` wrapper (or document `SSEStreamConsumer` as the sanctioned client with a typed entry point).
- [ ] Extend `web_ui/src/lib/api/types.ts` with `history?: { role: string; content: string }[]` on the request type used for `/ask` and `/ask/stream`, matching Python's `QuestionRequest.history`.
- [ ] Write `contracts/api.openapi.yaml` (OpenAPI 3.1) covering every route enumerated above. Reserve (do not implement) `grounding: "grounded"|"general"` on `QuestionResponse`/the stream `done` payload and `learn: [{slide_id, title, section, score, reason}]` as forward-compatible optional fields for #72 and #82.
- [ ] Write `contracts/tests/` — a backend-agnostic conformance suite parameterized on `base_url`, covering every route including a real `/ask/stream` round trip with a token, a mid-stream cancellation, and an error path.
- [ ] Wire `contracts/tests/` into CI running against `api_server.py`.
- [ ] Short `docs/adr/` note recording the loopback-default decision and the reserved `grounding`/`learn` fields.

Out of scope: implementing the Node backend or sidecar (A7/#57, WS-B), performance/benchmarking (A2/#52), bearer-token/loopback transport hardening beyond the bind-address default (owned by WS-B PR 2/#60), any renderer surface other than the chat streaming client and shared types.

## Acceptance
| Case | Required proof |
|---|---|
| `/ask/stream` completes end-to-end against a real engine | Integration test in `contracts/tests/` receives a `token` event then a terminal `done` event with `sources`/`context_length`/`inference_time` |
| Client disconnect cancels in-flight work | Test aborts mid-stream, asserts the engine's cancellation path is invoked |
| `contracts/tests` conformance suite is green | `pytest contracts/tests --base-url=http://127.0.0.1:8080` exits 0 |
| CORS allows the `/settings` PUT | Test asserts `PUT` in the CORS `allow_methods` header for an OPTIONS preflight |
| Default bind is loopback-only | Test with no `API_HOST` set asserts host resolves to `127.0.0.1` |
| Renderer server-mode chat works against fixed Python backend | Manual smoke test + existing Playwright chat spec in `server` mode |
| Regression: `/ask` still works unchanged | Existing `tests/test_api_asyncio_thread.py` remains green |

## Evidence and exit gate
Closing PR must attach: the new/updated test file names (e.g. `tests/test_api_ask_stream.py`, `contracts/tests/test_conformance.py`), a CI run link showing the conformance job green, `contracts/api.openapi.yaml` diff, and the manual smoke-test transcript for server-mode chat. Evidence is invalidated if the OpenAPI contract changes after this PR merges without a corresponding conformance-suite update.

| Slot | Issue | Work |
|---|---|---|

| Slot | Issue | Work |
|---|---|---|

| Slot | Issue | Work |
|---|---|---|

| Slot | Issue | Work |
|---|---|---|

| Slot | Issue | Work |
|---|---|---|

zaxbysauce:
Closes #51

[WS-A PR 1/8] Repair and freeze the renderer/Python API contract (fix /ask/stream, SSE lifecycle, cancel, CORS, loopback default)

Part of epic #50 (Trainingapp v3). Trace artifacts: `.agents/issue-traces/51-repair-renderer-python-api-contract/` (protocol 3.0.0, tier L, all gates APPROVE).

## Root Cause

`POST /ask/stream` (api_server.py:722 pre-fix) assigned the unawaited `asyncio.to_thread(...)` coroutine to `query_future` and then called `.done()` (line 736) and `.result()` (line 745) on that coroutine object — `AttributeError: 'coroutine' object has no attribute 'done'`, swallowed by the broad except arm into a single SSE `error` event. The RAG query never executed (`engine.query invocations: 0` in the frozen base log) and clients received zero tokens over an HTTP 200 `text/event-stream`. It survived because the only behavioral /ask/stream tests lived in root-level `test_api_endpoints.py`, which pytest never collects (`testpaths = tests tests/integration`; CI runs `pytest tests/`). Four contract-surface defects compounded it: cancellation_event never passed to `engine.query` (so the existing `QueryCancelled` arm was dead code and client disconnects burned CPU), the cancelled-done payload omitted `context_length` (renderer done-detection at web_ui/src/lib/api/streaming.ts:369 requires `sources` AND `context_length`), `API_HOST` defaulted to `0.0.0.0` with auth off by default, and CORS `allow_methods` omitted `PUT` while `PUT /settings` is part of the renderer surface.

## Fix

- api_server.py `ask_question_stream`: schedule via `asyncio.create_task(asyncio.to_thread(...))` with a retained task reference; queue-race loop (`asyncio.wait FIRST_COMPLETED` over `queue.get()` and the query task) yields each token as it arrives, drains stragglers, then emits exactly one `done` terminal (sources + context_length + inference_time) or one `error` terminal.
- Cancellation: per-request `threading.Event` passed to `engine.query` (which already supported it, rag_engine.py:317); set in the generator's `finally` on any exit (client disconnect / cancel / error), abandoned task reaped; `except QueryCancelled` emits a renderer-detectable done payload with `cancelled: true` and `context_length: 0` under a `stream_open` guard.
- `API_HOST` defaults to `127.0.0.1` (breaking-ish default flip; explicit `API_HOST=0.0.0.0` opt-in documented in CHANGELOG).
- CORS `allow_methods=["GET","POST","PUT","DELETE"]`.
- New unauthenticated `GET /health` -> `{"status":"ok","engine_ready":bool}` registered before the static archive mount.
- New `contracts/api.openapi.yaml` (OpenAPI 3.1): all 12 routes incl. auth/status, auth/token, ask, ask/stream (SSE semantics documented), ingest, ingest/file, ingest/batch, documents GET+DELETE, search, settings GET+PUT, stats, health; `learn[]` and `grounding` RESERVED for WS-C #72.
- New `contracts/tests/run_conformance.py`: backend-agnostic conformance suite (`--base-url` for any backend, `--asgi api_server:app` for in-process with a stub engine; `--safe` skips destructive routes) with a bidirectional spec-vs-app route drift check.
- CI: new fast `conformance` job in .github/workflows/test.yml runs `scripts/check_test_collection.py` then the conformance runner.
- Defect-class guardrail `scripts/check_test_collection.py`; the three never-collected root-level test files moved under `tests/` and refreshed; new `tests/test_api_ask_stream.py` regression tests; USAGE.md streaming docs aligned with the real SSE contract; CHANGELOG entry.

- Defect class: fire-and-forget coroutine/future misuse at an async call site, hidden by tests CI never collects.
- Sweep result: P1 (unawaited to_thread assignments) 0 remaining; P2 (`.done()`/`.result()` misuse) 5 hits, all on genuine Tasks in the fixed handler; P3 (orphan test files) 3 hits, all moved+refreshed (5 stale BM25 sub-tests skipped with documented revival reason); P4 (contract drift) enforced bidirectionally; P5 (dead except-arm on unwired parameter) 1 hit, wired and now exercised.
- Guardrail: `scripts/check_test_collection.py` (CI-enforced) — demonstrated RED on the pre-fix layout (3 orphans listed, exit 1) and GREEN after disposition; contract drift check demonstrated RED in both directions (spec path removed -> FAIL via extra_in_app; spec-only path -> FAIL via missing_from_app).

## Tests

- Frozen acceptance checks (10, red-checkpointed before the fix): `repro-check.sh run` per check — C1-C4, C7, C10 DISCRIMINATING RED->GREEN; C5/C6 NEW-SURFACE ERROR->GREEN; C8/C9 PRESERVING GREEN->GREEN. Logs: `.agents/issue-traces/51-repair-renderer-python-api-contract/phase-4-runs.log`, `repro/C*.base.log`/`C*.head.log`.
- Regression test: `pytest tests/test_api_ask_stream.py` -> 6 passed
- Impacted suite: `pytest tests/test_api_ask_stream.py tests/test_api_endpoints.py tests/test_api_asyncio_thread.py tests/test_cancellation_gui.py tests/test_vector_store_bm25.py tests/test_api_server_ragconfig_wiring.py tests/test_api_validation.py tests/test_api_server_ingest_file_doubleread.py tests/test_api_server_ingest_file_xlsx.py` -> 189 passed, 5 skipped (documented stale BM25 fixtures, see 08a)
- Conformance: `python contracts/tests/run_conformance.py --asgi api_server:app` -> PASS 12/12
- Lint/format: pre-commit (black, isort, flake8 --max-line-length=100, yaml checks) -> all Passed
- Deferred-work scan: `scan-deferred.sh` -> clean (base=origin/master)

## Regression Protection

- `tests/test_api_ask_stream.py`: token streaming + single renderer-contract done terminal; cancelled terminal carries context_length; exactly-one error terminal; /health; CORS PUT preflight; loopback API_HOST default.
- `tests/test_api_ask_stream.py::test_ask_stream_engine_error...` negative/adversarial: engine RuntimeError yields exactly one error terminal and no trailing tokens.
- The moved/refreshed suites re-joined CI collection; `scripts/check_test_collection.py` (now CI-enforced) prevents silent future orphans; conformance drift check prevents spec/app divergence in both directions.

| Acceptance criterion | Evidence |
|---|---|
| AC1 /ask/stream executes + streams + done terminal | C1 RED->GREEN (`CHECK C1: PASS`); base RED with `engine.query invocations: 0` |
| AC2 error terminal exactly-once | C9 GREEN at base and head (`CHECK C9: PASS`) |
| AC3 disconnect -> cancellation_event | C2 RED->GREEN (`CHECK C2: PASS`; real uvicorn + real client abort) |
| AC4 cancelled payload renderer-detectable | C7 RED->GREEN (`CHECK C7: PASS`) |
| AC5 loopback default | C3 RED->GREEN (`CHECK C3: PASS`) |
| AC6 CORS PUT | C4 RED->GREEN (`CHECK C4: PASS`) |
| AC7 frozen OpenAPI contract | C5 ERROR->GREEN (`CHECK C5: PASS`) |
| AC8 conformance suite green on Python | C6 ERROR->GREEN (`CONFORMANCE: PASS 12/12`) |
| AC9 CI runs conformance | C10 RED->GREEN (`CHECK C10: PASS`) |
| AC10 renderer server-mode manual+Playwright | NON-EXECUTABLE (EXTERNAL_SERVICE_UNAVAILABLE: model weights operator-acquired, absent in env/CI). Substitute: C1/C2/C4/C7 wire-contract checks + manual procedure below |
| AC11 /ask + history preserved | C8 GREEN->GREEN (`CHECK C8: PASS`) |

Manual server-mode smoke procedure (AC10 substitute, operator-run): start `python api_server.py`, in the web UI Settings choose server mode + `http://127.0.0.1:8080`, ask a question in Chat, verify tokens stream and the done state renders sources; click Stop mid-answer and verify generation stops server-side (CPU drops) with a clean cancelled state.

## Invariant Audit

Repository has no dedicated invariants/architecture-contract doc — none documented. Contract-surface invariants touched and pinned by tests: SSE terminal semantics (exactly one done XOR error — C9/C1), renderer done-detection payload keys (C1/C7), `/ask` history threading (C8, issue #40 behavior), route/method surface (conformance drift check).

## Risk and Rollback

- Risk level: medium (one intentional breaking-ish default: API_HOST now 127.0.0.1; LAN deployments must set `API_HOST=0.0.0.0` explicitly and are encouraged to enable `ENABLE_AUTH=true`)
- Rollback: single-commit revert (no migrations, no persisted-state changes)
- Residual risk: success-path wire format, error format, and the /ask path are unchanged (C8/C9 pin them); 5 stale BM25 sub-tests remain skipped with documented revival reason

## Waivers (or none)

none

## Merge status

PR head: 4bd6240caddff67e745a9c581cdd2b7fc5baf74e1

## Issue Closure

Closes #51

<a href="https://cubic.dev/pr/ZaxbyHub/trainingapp/pull/90?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Feedback round (swarm-pr-feedback, 2026-09-05)

All review findings resolved in ba90e57: PRR-001 (collection ImportError), PRR-003/004 (sse-starlette + PyYAML declared), PRR-007 (cancelled-done terminal now reachable with the real engine via the CANCELLED_ANSWER sentinel), PRR-008/012 (StreamEvent oneOf schema), PRR-009/013 (docs drift incl. CHANGELOG rename citation + INSTALL.md API_HOST), PRR-010/011 (conformance destructive opt-in, 5MB SSE cap, malformed-payload surfacing), plus the Copilot thread (httpx.AsyncHTTPTransport in --base-url mode). All 5 PR review threads answered and resolved. Gates: reviewer APPROVE + final critic APPROVE on the committed tree. Validation: flake8 clean, collection guardrail OK, focused pytest green, conformance 12/12 with bidirectional drift empty.

CI follow-up (4bd6240): the full-suite matrix run surfaced two more live consumers of the removed re-exports (tests/test_phase1_fixes.py::test_validate_url, tests/test_config.py::test_api_server_imports_settings) — both repaired; all other consumers are module-skipped per the PRR-002 precedent.


zaxbysauce (commented):
## Swarm PR review — /swarm-pr-review 90

**PR intent:** Repair `POST /ask/stream` (unawaited-coroutine bug that silently produced zero-token SSE streams), add real cancellation support, flip `API_HOST` default to loopback, and freeze a renderer/Python API contract via a new `contracts/api.openapi.yaml` + bidirectional drift checker.

**Method:** 5 parallel adversarial explorer lanes (correctness, security, dependency/CI, docs/intent-vs-actual, tests/falsifiability), each told to assume the work is wrong until proven otherwise. I then independently re-verified every candidate against the actual PR-branch source before reporting.

### Confirmed findings (real, evidence-checked)

1. **Cancellation never reaches the client — the PR's headline cancellation feature is dead in the primary path.** `api_server.py:794` (`except QueryCancelled`) is unreachable in practice: `RAGEngine.query` (`rag_engine.py:373-417, 591-624`) catches `QueryCancelled` **internally** at every checkpoint and returns a normal `QueryResult(answer="[Cancelled]", sources=[], context_length=0, ...)` instead of re-raising it. Since `api_server.py:778` (`result = query_task.result()`) never sees the exception, execution always falls through the success path (`api_server.py:783-793`), emitting `{"done": true, "sources": [], "context_length": 0, ...}` **without** `cancelled: true`. The client can't tell "cancelled" from "answered with zero context." `tests/test_api_ask_stream.py`'s cancellation test only passes because it mocks `Engine.query` to `raise QueryCancelled()` directly — that's not how the real engine behaves, so the test gives false confidence. This is the one thing I'd block on: either make `RAGEngine.query` re-raise (or return a `cancelled` flag on `QueryResult` that `api_server.py` checks), and add a test using the real `RAGEngine`/checkpoints, not a raise-happy stub.

2. **`contracts/api.openapi.yaml`'s own `StreamEvent` schema doesn't match what the server emits.** It declares `required: [variant]` with `variant: enum [token, done, error]` (`contracts/api.openapi.yaml:435-441`), but `api_server.py` never emits a `variant` key anywhere (grep confirms zero occurrences) — real payloads are bare `{"token": ...}` / `{"done": ..., "sources": ...}` / `{"error": ...}`. Since this contract is explicitly documented as authoritative for the future WS-B Electron host (#61), an implementer coding strictly to the schema would build an incompatible client. Not caught by `run_conformance.py` (it does ad-hoc dict-key checks, not schema validation).

### Real, non-blocking gaps

3. Client-disconnect cancellation (`api_server.py:819-834`, the `finally` block that sets `cancellation_event`/cancels the task) is never exercised by a real disconnect in tests — only the mocked-raise path is tested. Ties directly to finding #1.
4. `scripts/check_test_collection.py` (the new guardrail meant to prevent "test exists but CI never collects it" from recurring) verifies file *location* against `pytest.ini`'s `testpaths`, not actual pytest collectability — an import error or a future `addopts`/`--ignore` exclusion would pass silently. It also has no test of its own detection logic.
5. `contracts/tests/run_conformance.py`'s `check_contract_drift` silently `skip()`s (not fail) when PyYAML is unimportable (`run_conformance.py:333-335`), and a skip never fails the job (`total = passed+failed`, skipped excluded). PyYAML isn't declared in `requirements.txt` or the `conformance` job's install step — it currently works only because `sentence-transformers → huggingface_hub` transitively pulls it in. If that chain ever changes, this safety net goes quiet with zero CI signal. Recommend declaring PyYAML explicitly and asserting the check wasn't skipped.
6. `stream_open` (`api_server.py:724`) is dead-weight: it's only ever set `False` in the `finally` block, which runs after both `except` arms already read it, so it can never actually be `False` at either read site. Not a bug, but the docstring's safety claim is misleading — worth a comment fix or removal.
7. `except Exception as e: logger.error("...: %s", e)` (`api_server.py:811`) still doesn't log a stack trace (no `exc_info=True`/`logger.exception`) — a recurrence of this exact bug class (silently-swallowed exception) would again be hard to diagnose from logs alone.
8. Minor: `QuestionRequest.question` has no whitespace-strip validation (unlike `/search`'s `validate_query_not_whitespace`), so `"   "` passes validation. Low severity, pre-existing inconsistency, not introduced by this PR's core fix.

### Pre-existing / explicitly out of scope
- `ENABLE_AUTH` still defaults to `false` (`auth.py`) — the `API_HOST` loopback default flip narrows exposure but doesn't touch this. Appears intentionally deferred (documented as a known residual), not silently dropped — worth confirming it's tracked separately.
- CORS, auth-coverage matrix, `/health` endpoint, route registration order, and the OpenAPI route-count claim (12 documented routes) were all checked and are clean/consistent.

### Verdict
**Do not merge as-is.** The core bug (#51's unawaited-coroutine crash) is genuinely fixed and well-tested for the happy/error paths. But the PR's second headline feature — cancellation — silently fails to deliver the `cancelled: true` contract it documents and tests claim to cover, because the exception it's designed to catch never actually reaches it. That's a real regression risk for the renderer teams building against this "frozen" contract. Recommend fixing #1 (and ideally #2) before merge; items 3-8 are good fast-follow or same-PR polish but not blockers on their own.

api_server.py
```diff
@@ -708,69 +716,122 @@ async def ask_question_stream(
...
+                    "data": json.dumps(
+                        {
+                            "done": True,
+                            "sources": sources,
+                            "context_length": context_length,
+                            "inference_time": inference_time,
+                        }
+                    ),
                 }
             except QueryCancelled:
```
zaxbysauce, line 822: CONFIRMED (blocking): this `except QueryCancelled` arm is unreachable in the normal cancellation flow. `RAGEngine.query` (rag_engine.py:373-417, 591-624) catches `QueryCancelled` internally at every checkpoint and returns a normal `QueryResult(answer="[Cancelled]", sources=[], context_length=0, ...)` rather than re-raising it — so `query_task.result()` at line 778 never raises, and execution always falls through the success path above (lines 783-793), emitting `done: true` **without** `cancelled: true`. The client cannot distinguish a cancelled query from one that completed with zero context. `tests/test_api_ask_stream.py`'s cancellation test only passes because it mocks `Engine.query` to `raise QueryCancelled()` directly, which doesn't reflect the real engine's swallow-and-return behavior — that's a false-green test. Suggest either making `RAGEngine.query` re-raise `QueryCancelled` when the cancellation event fired, or having it set a `cancelled` flag on `QueryResult` that this handler checks.

contracts/api.openapi.yaml
```diff
@@ -0,0 +1,616 @@
...
+          type: array
+          description: RESERVED for WS-C #72 (OpMed learning links). Not emitted yet.
+          items:
+            type: object
+        grounding:
+          type: object
+          description: RESERVED for WS-C #72 (grounded/general provenance). Not emitted yet.
+          additionalProperties: true
+    StreamEvent:
+      type: object
```
zaxbysauce: CONFIRMED: `StreamEvent` declares `required: [variant]` with `variant: enum [token, done, error]`, but `api_server.py`'s SSE generator never emits a `variant` key anywhere (grep confirms zero occurrences) — actual payloads are bare `{"token": ...}` / `{"done": ..., "sources": ...}` / `{"error": ...}`. Since this spec is documented as authoritative for the future WS-B Electron host (#61), a consumer coding strictly to this schema would build a wire-incompatible client. `contracts/tests/run_conformance.py` does ad-hoc dict-key checks rather than JSON-Schema validation against `components.schemas`, so it doesn't catch this. Please either drop `variant` from the schema (matching actual behavior) or add it to the emitted payloads and cover it in the conformance check.

api_server.py
zaxbysauce: Resolved in ba90e57 (PRR-007). rag_engine converts cancellation into a normal QueryResult instead of re-raising, so the success path now detects it: `if cancellation_event.is_set() or result.answer == CANCELLED_ANSWER:` emits the cancelled-done terminal carrying sources+context_length (renderer done-detection contract). The `except QueryCancelled` arm stays as defense-in-depth and is exercised by conformance's CancelEngine. Test stub mirrors real-engine semantics (returns the sentinel, does NOT set the event) — validated by independent reviewer + critic rounds including a revert-probe.

contracts/api.openapi.yaml
zaxbysauce: Resolved in ba90e57 (PRR-008): StreamEvent reshaped to a oneOf of TokenEvent/DoneEvent/ErrorEvent payload schemas; the unenforced `variant` requirement is gone and DoneEvent requires done+sources+context_length exactly as api_server emits. contract_drift passes with missing_from_app=[] extra_in_app=[].

.github/workflows/test.yml
```diff
@@ -12,34 +12,64 @@ jobs:
...
+          path: ~\AppData\Local\pip\Cache
+          key: ${{ runner.os }}-conformance-${{ hashFiles('**/requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-conformance-
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          pip install httpx
```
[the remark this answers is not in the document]
zaxbysauce: Resolved in ba90e57 (PRR-003/004): requirements.txt now declares sse-starlette>=2.0.0,<4.0.0 (and PyYAML for the drift check), so HAS_SSE is true on a clean install and the CI conformance job sees /ask/stream.

contracts/api.openapi.yaml
```diff
@@ -0,0 +1,616 @@
...
+      description: One SSE data payload (JSON) on event "message" or "error"
+      required: [variant]
+      properties:
+        variant:
+          type: string
+          enum: [token, done, error]
```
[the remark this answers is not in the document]
zaxbysauce: Resolved in ba90e57 (PRR-008): same reshape as the sibling thread — `variant` requirement removed; StreamEvent is now a oneOf of the three payload shapes that actually appear on the wire.

PR #96, opened 2026-09-08
feat(desktop): secure the Electron protocol and loopback API transport (issue #60)

### LINKED ISSUE #60: [Workstream B] PR 2 of 9: Secure the Electron protocol and loopback API transport
**Roadmap:** B2 · **Workstream B: Electron shell and desktop backend**
**Sequence:** PR 2 of 9 (tracking slot; dependencies define execution order)

## Problem
The Python backend that must interoperate with the Electron transport (per ADR-0003, #57) binds `0.0.0.0` by default with auth off (`api_server.py:1002` default `API_HOST="0.0.0.0"`, `auth.py:25` default `ENABLE_AUTH=False`; CONFIRMED per `audit_report.md` gap #15 and `critic_sol_verdict.md` risk 7 — re-verify exact lines at implementation time, this issue's file scope is `desktop/`). No equivalent transport-hardening design exists for Electron yet, so a naive reuse of this contract ships a loopback RAG endpoint any local process could call.

## Requested change
Bind whatever backend B3/#61 hosts to `127.0.0.1` on a random free port, mint a per-launch unguessable bearer token in the main process, deliver it to the renderer ONLY via `contextBridge`/IPC (never localStorage, never a URL param), and lock down navigation/window-open and CSP so the only reachable origins are `app://` and the loopback host:port.

- [web_ui/src/lib/api/index.ts:40](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/web_ui/src/lib/api/index.ts#L40) — CONFIRMED: `export const apiClient = new ApiClient('');` — this is the client B9's real callers will feed with the token/origin plumbing this issue builds; do not change its constructor signature if avoidable.
- `web_ui/scripts/start.ps1:165-171` — CONFIRMED, the existing airgapped server already treats header discipline as load-bearing (every response gets COOP/COEP/CORP); this issue's CSP policy for `app://` responses should follow the same "every response, including errors" discipline.
- No token-delivery, origin-check, or CSP code exists yet (`desktop/` is created fresh in B1/#59). The Python `0.0.0.0`/auth-off defaults above are the negative example this issue must not repeat regardless of which side of ADR-0003 wins.

#59 (B1) supplies the shell, main/preload split, and `app://` handler this issue hardens. Owns: token minting/delivery, origin/host allow-list, CSP policy, navigation/`window.open` lockdown, `docs/security/desktop.md`. Does NOT own: which process serves `/ask/stream` (B3/#61's ADR-0003 choice) — checks must be backend-agnostic. Not parallel-safe with B3-B9 (all build on this transport); land immediately after B1.

## Required scope
- [ ] `desktop/main/security/token.ts` — `crypto.randomBytes(32).toString('hex')` per launch, main-process memory only, never persisted/logged.
- [ ] `desktop/preload/index.ts` extension — `contextBridge.exposeInMainWorld('desktopApi', { getAuthToken: () => ipcRenderer.invoke('desktop:get-token') })`.
- [ ] `desktop/main/security/loopback-guard.ts` — reject requests whose `Origin`/`Host` mismatches `app://`/`127.0.0.1:<port>`, or whose bearer token is missing/wrong, before reaching the backend chosen by B3.
- [ ] Config keys: `security.tokenHeaderName` (default `"X-Desktop-Token"`), `security.allowedOrigins` (default `["app://*"]`; dev-only additions gated by a build-time flag).
- [ ] `desktop/main/security/csp.ts` — strict CSP (`default-src 'self' app:; connect-src 'self' app: http://127.0.0.1:*; script-src 'self'; object-src 'none'`), tuned against actual WASM needs, relaxations documented inline.
- [ ] Navigation lockdown in `desktop/main/index.ts`: `will-navigate` denies non-`app://` targets; `setWindowOpenHandler(() => ({ action: 'deny' }))`.
- [ ] `docs/security/desktop.md` — threat model: token theft, malicious renderer content, LAN exposure, token replay across restarts.
- [ ] Test file exercising every rejection path.
- Out of scope: the backend process itself (B3); rate limiting beyond origin+token; TLS (loopback-only).

## Acceptance
| Case | Required proof |
| --- | --- |
| Missing token rejected | Test: request without token header rejected before backend logic runs |
| Wrong-origin rejected | Test: `Origin: http://evil.example` rejected regardless of token validity |
| Token never in localStorage | Test/grep: no `localStorage.setItem` call ever writes the token |
| External navigation blocked | Test: `will-navigate`/`window.open` to external URLs denied |
| CSP effective | Manual DevTools check + a blocked inline-script injection attempt |
| Threat model documented | `docs/security/desktop.md` merged and reviewed |

## Evidence and exit gate
Attach: test paths, `docs/security/desktop.md`, and a manual verification transcript (401 on a missing-token request against the packaged app on the reference laptop). Invalidated if ADR-0003 selects the sidecar path and its own auth/CORS defaults are left unhardened — this guard must sit in front of the sidecar too.

| Slot | Issue | Work |
|---|---|---|

| Slot | Issue | Work |
|---|---|---|

| Slot | Issue | Work |
|---|---|---|

| Slot | Issue | Work |
|---|---|---|

| Slot | Issue | Work |
|---|---|---|

zaxbysauce:
Closes #60

## Root Cause

The Electron desktop shell shipped by Workstream B1 (PR #95) deliberately deferred all transport hardening to B2 (`desktop/main/index.ts:6-8`, `desktop/main/protocol.ts:9-10`, `desktop/README.md:50-66`): `app://` responses carried no CSP (and no COOP/COEP/CORP), the window had no `will-navigate`/`window.open` lockdown, the preload exposed no token bridge, and no request gate existed in front of the future loopback backend — while `auth.py:25` still defaults `ENABLE_AUTH=false`, meaning an unguarded loopback port would be callable by any local process. This PR builds that missing layer (Workstream B2).

## Fix

Minimal layer at the right trust boundary (the main process; the renderer is untrusted input):

- `desktop/main/security/` (new): `config.ts` (`security.tokenHeaderName` default `X-Desktop-Token`, `security.allowedOrigins` default `["app://*"]`; dev-only origin additions require `!app.isPackaged` AND an explicit `TRAININGAPP_DESKTOP_DEV_ORIGINS` env opt-in, so packaged builds categorically cannot gain origins); `token.ts` (per-launch `crypto.randomBytes(32).toString('hex')`, main-process memory only, never persisted/logged); `loopback-guard.ts` (backend-agnostic request gate: origin allow-list incl. `Origin: null`/empty and subdomain tricks rejected 403 regardless of token, literal-loopback-only Host gate — `localhost` names and LAN IPs rejected as DNS-rebinding hardening — then token equality check with static 401 bodies that never echo the token); `csp.ts` (strict CSP with inline-documented relaxations: `'wasm-unsafe-eval'` for ONNX/wllama, a sha256 pin for the ONE legitimate inline theme-bootstrap script, `worker-src blob:` for ORT threading, loopback `connect-src`); `index.ts` barrel + per-launch guard holder.
- `desktop/main/index.ts`: navigation lockdown (`will-navigate` allows `app://` only — plus the dev-server origin in dev mode; `webContents.setWindowOpenHandler` denies every `window.open`; note this is the only surface that exists at runtime in Electron 44 — probe-verified), and `bootstrap()` wires config → token → `ipcMain.handle('desktop:get-token')` → guard construction with a fail-fast `app.quit()` if the guard was not constructed.
- `desktop/main/protocol.ts`: every `app://` response (success, 403, 404) carries CSP + COOP/COEP/CORP — the "every response, including errors" discipline of `web_ui/scripts/start.ps1`; COOP/COEP also enable `crossOriginIsolated` so ONNX can use `numThreads > 1` under `app://`.
- `desktop/preload/index.ts`: additive `contextBridge.exposeInMainWorld('desktopApi', { getAuthToken: () => ipcRenderer.invoke('desktop:get-token') })`; the legacy `trainingapp` namespace is unchanged; preload stays CommonJS (`.cjs`, sandboxed-renderer requirement, guarded by `verify-preload-format.mjs`).
- `docs/security/desktop.md` (new): threat model (token theft, malicious renderer content, LAN exposure, token replay across restarts) and the **binding B3 integration contract**: #61's backend MUST bind `127.0.0.1` on a random free port, MUST mount `getLoopbackGuard()` in front of every route before backend logic, MUST read the header name from config, MUST NOT log the token. This is the issue's own ownership split, enforced by the fail-fast bootstrap assertion — not silent deferral.
- `.gitattributes`: `web_ui/index.html` pinned to LF so the CSP sha256 pin is identical on every checkout (the pin would otherwise be autocrlf-dependent).
- `web_ui/` itself: byte-zero diff (B9/#67 owns renderer consumption; the `ApiClient('')` constructor is untouched as the issue requires).

Reviewer wiring note (for anyone checking): the new `desktop/main/security/*.ts` files need no tsconfig/CI changes — the existing `tsconfig.json` include `main/**/*.ts` compiles them and `desktop-build.yml`'s existing `npm --prefix desktop test` picks up the new specs.

- Defect class: a new transport/process seam ships without its trust-boundary enforcement (no origin allow-list, no auth gate, no CSP, no navigation lockdown).
- Sweep result: 11 hits across 7 predicates (window-open surfaces, navigation policy, exposed bridges, response construction, bind addresses, token storage, webPreferences coverage) — every hit dispositioned as the enforcement itself, test infrastructure, or a documented out-of-class surface (browser-mode JWT in `web_ui/src/lib/api/auth.ts` is a different token for a different transport; relationship documented in `docs/security/desktop.md`).
- Guardrail: regression family — 7 frozen acceptance specs + `desktop/repro/check-b2.sh` run in CI via the existing `desktop-build.yml` test step — plus the documented "every seam threads the security barrel" invariant. Demonstrated to bite with three mutation probes (nav allow-list neutered → `FAIL navigation-lockdown`; `wasm-unsafe-eval` dropped → `FAIL csp`; token check neutered → `FAIL loopback-guard`; each restored → PASS) and by the frozen red checkpoint (base logs show every check RED/ERROR at 8af74d4).

## Tests

- Regression test: `npm --prefix desktop test` → 15 files / 94 tests PASS (B1 files unchanged except the intentional two-namespace preload contract update in `secure-defaults.test.ts`; 7 frozen B2 specs + feedback-round additive specs: guard edge cases, CI-enforced token-storage greps, security-header pinning, pin-drift guard).
- Acceptance checks (issue-tracer red checkpoint): all 7 ids `DESKTOP-60 CHECK: PASS` via `repro-check.sh run` — base 8af74d4 RED/ERROR for the expected reason, head 6139555 GREEN; `verify-checkpoint` OK on all 9 manifest paths.
- Lint/type/build/security checks: `npm --prefix desktop run compile` → PASS; `node desktop/scripts/verify-preload-format.mjs` → OK; real-Electron runtime smoke `electron scripts/preload-runtime-smoke.cjs` → OK (exit 0); manual verification transcript against the COMPILED guard over real loopback HTTP → 10/10 cases as expected (missing/empty/wrong token → 401; wrong/null origin, localhost-name, LAN-IP hosts → 403 even with a valid token; valid origin+token → backend).
- Deferred-work scan: clean (no TODO/FIXME/stub markers on added lines).

## Regression Protection

- `desktop/src/__tests__/b2-loopback-guard.test.ts` (15) + `b2-loopback-origin.test.ts` (8): every rejection path couples the status assertion with `delegate not called` (backend isolation); pass-through positive controls included.
- `b2-csp.test.ts` (4): CSP on 200/403/404 + a negative control proving the strictness parser rejects permissive policies.
- `b2-token-lifecycle.test.ts` (4): fs-write and console spies; 64-hex format; per-mint uniqueness.
- `b2-security-config.test.ts` (6): defaults + packaged-refusal matrix.
- `b2-csp-pin.test.ts` (2): CSP pin vs the tracked `web_ui/index.html` inline script hash (drift fails CI) + never-unsafe-inline.
- `b2-token-bridge.test.ts` (3): bridge shape + IPC routing + legacy namespace compat.
- Test drift review: `secure-defaults.test.ts` single-namespace assertion updated to the two-namespace contract — intentional, AC-driven, both shapes still pinned (judged "not a weakening" by the independent implementation review).

|---|---|
| AC1 missing token rejected before backend logic | C1 `loopback-guard` PASS; manual transcript: missing/empty/wrong token → 401, delegate untouched |
| AC2 wrong origin rejected regardless of token | C7 `loopback-origin` PASS (evil.example/'null'/subdomain/LAN/localhost-name with valid token → 403); mutation probe bites |
| AC3 token only via contextBridge/IPC; never storage/URL | C2 `token-bridge` PASS + driver structural greps (zero hits) + C5 fs/console spies |
| AC4 external navigation blocked | C3 `navigation-lockdown` PASS (will-navigate app://-only; window.open denied for every URL) |
| AC5 strict CSP on every response incl. errors; injection blocked | C4 `csp` PASS (200/403/404; negative control); pin-drift guard + built-dist pin check; relaxations documented inline in `csp.ts`. The issue's "Manual DevTools check" half is superseded by the automated header/injection assertions (same evidence a DevTools inspection would show) — packaged verification on the reference laptop is the E3/#86 residual |
| AC6 threat model documented | `docs/security/desktop.md` (four issue-named threats + B3 contract) |
| AC7 token lifecycle (randomBytes(32) hex, memory-only, never persisted/logged) | C5 `token-lifecycle` PASS; `token.ts` |
| AC8 config keys + packaged-build origin gate | C6 `security-config` PASS (defaults; packaged+env-opt-in refused) |

## Invariant Audit

No formal invariants/architecture-contract doc exists in the repo ("none documented"); the desktop-local contract (`desktop/README.md` security posture + `ARCHITECTURE.md` has no invariant section) was audited instead:

- B1 secure defaults (`contextIsolation`/`sandbox`/`nodeIntegration`/`webviewTag`): not touched — still pinned by `secure-defaults.test.ts` (negative-control spec) — 4/4 pass.
- Single-instance + bootstrap wiring order: not touched — `review-hardening.test.ts` 7/7, `single-instance.test.ts` 4/4.
- `app://` path-traversal safety: not touched (header wrapping only) — `app-protocol.test.ts` 18/18.
- Preload CommonJS/sandbox constraint (PRR95-001): preserved — `verify-preload-format.mjs` OK + real-Electron runtime smoke OK.
- Minimal default set: build OK (compile), typecheck OK (tsc via compile), tests 79/79, no generated artifacts staged (`desktop/dist` and `renderer/` are gitignored), push-protection secret scan clean.

## Risk and Rollback

- Risk level: low (additive main-process layer; renderer byte-zero; no schema/API/persistence change). Residual: if the strict CSP blocks an undocumented renderer capability, the documented fallback is a conscious `csp.ts` policy edit (pin-drift test will point at it); COOP/COEP assume the fully self-hosted renderer (no cross-origin subresources exist).
- Rollback: single-commit revert (`git revert 6139555`); no migrations or config state.

## Waivers (or none)

none

## Merge status

PR head: 6139555d9db7b1747ae37324b7330515731ed32c

## Issue Closure

Closes #60

## Feedback round (post-review fixes)

A swarm-pr-review pass (6 base lanes + 11 micro risk families) plus an independent 5-lane PR review produced findings resolved in commit c004204: B3 adapter contract corrections (absolute-URL requirement, case-insensitive headers, reserved `method`/preflight-after-gates), IPv6 and edge-case guard tests, COOP/COEP/CORP header pinning, built-dist CSP pin check (which caught a stale CRLF dist immediately), token-storage greps ported into CI vitest, test-reset hooks, and README corrections. Full closure ledger: `.zcode/pr-review/pr96-review-20260908/closure-ledger.md` (in-repo working notes; 11 rejected candidates listed with reasons). Independent reviewer and final critic both APPROVE the fix diff at c004204.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# CSP pin stability (issue #60): desktop/main/security/csp.ts hash-pins the inline
+# theme-bootstrap script of this file, so its served bytes must be identical on
+# every checkout regardless of core.autocrlf. LF everywhere, no exceptions.
+web_ui/index.html text eol=lf
--- a/desktop/README.md
+++ b/desktop/README.md
@@ -30,7 +30,9 @@ desktop/
                          `desktopApi.getAuthToken` (B2 token bridge, IPC-only)
   src/__tests__/         frozen acceptance specs (vitest, stubbed electron)
   test/electron-stub.ts  in-memory electron for the specs
-  repro/check.sh         acceptance-check driver (C1..C6)
+  repro/check.sh         acceptance-check driver for issue #59 (C1..C6)
+  repro/check-b2.sh      acceptance-check driver for issue #60 (B2 specs)
+  scripts/               preload format guard + runtime smoke (CI)
   electron-builder.yml   unsigned NSIS x64 packaging (appId com.zaxbyhub.trainingapp)
   renderer/              build-time staging of web_ui/dist (gitignored)
   desktop-release/       electron-builder output (gitignored)
@@ -40,8 +42,8 @@ desktop/
 
 | Script | What it does |
 |---|---|
-| `npm run compile` | tsc -> `dist/` (main + preload, ESM, nodenext) |
-| `npm test` | all three acceptance spec files (vitest) |
+| `npm run compile` | tsc -> `dist/` (main as ESM/nodenext; preload as CommonJS, renamed .cjs) |
+| `npm test` | all acceptance spec files (vitest; B1 + B2 suites) |
 | `npm run dev` | compile + launch Electron with `--dev` (expects the vite dev server already running — use `desktop:dev` to start both) |
 | `npm run desktop:dev` | vite dev (web_ui) + Electron with reload, via `concurrently` |
 | `npm run desktop:build` | build web_ui, copy `web_ui/dist` -> `renderer/`, compile, run electron-builder (NSIS x64, unsigned) |
@@ -51,6 +53,9 @@ Dev mode: Electron loads `http://localhost:5173` when `--dev` is in argv, or
 whatever URL `ELECTRON_START_URL` points at. Production: `app://index.html`
 served from `<resourcesPath>/web_ui`. Note the packaged binary does not start a
 vite server — `--dev` on an installed app shows nothing; use `desktop:dev`.
+In dev mode the `app://` protocol handler is intentionally not registered (the
+vite server serves the renderer), so an `app://` navigation — allowed by the
+navigation policy — fails to load; that combination is production-only.
 
 ## Security posture (baseline, per issue #59)
 
--- a/desktop/README.md
+++ b/desktop/README.md
@@ -20,9 +20,14 @@ per-package lockfiles is the honest migration, not a blanket hoist).
 
 ```
 desktop/
-  main/index.ts          app bootstrap: single-instance lock, window, secure defaults
-  main/protocol.ts       app:// file protocol (traversal-refusing, MIME-mapped)
-  preload/index.ts       empty contextBridge stub (B2/#60 fills it)
+  main/index.ts          app bootstrap: single-instance lock, window, secure defaults,
+                         transport-security wiring, navigation lockdown (B2)
+  main/protocol.ts       app:// file protocol (traversal-refusing, MIME-mapped,
+                         CSP + COOP/COEP/CORP on every response, B2)
+  main/security/         B2 transport security: config, per-launch token,
+                         backend-agnostic loopback guard, CSP policy (barrel: index.ts)
+  preload/index.ts       contextBridge namespaces: `trainingapp` (shell) +
+                         `desktopApi.getAuthToken` (B2 token bridge, IPC-only)
   src/__tests__/         frozen acceptance specs (vitest, stubbed electron)
   test/electron-stub.ts  in-memory electron for the specs
   repro/check.sh         acceptance-check driver (C1..C6)
@@ -60,10 +65,25 @@ vite server — `--dev` on an installed app shows nothing; use `desktop:dev`.
   renderer root, re-checks containment through `realpath` (symlink escape),
   and returns 403/404 rather than reading anything outside the root
   (spec-pinned including Windows drive-letter and `%00` cases).
-- Preload exposes a single empty namespace (`window.trainingapp`) and nothing
-  else.
+- Preload exposes two non-privileged namespaces (`window.trainingapp`,
+  `window.desktopApi.getAuthToken`) and nothing else; the per-launch transport
+  token reaches the renderer only through that bridge (never storage, never a
+  URL).
+- B2 transport security (issue #60): per-launch token
+  (`crypto.randomBytes(32)` hex, main-process memory only), strict CSP +
+  COOP/COEP/CORP on every app:// response (relaxations documented inline in
+  `desktop/main/security/csp.ts`), navigation/`window.open` lockdown, and a
+  backend-agnostic loopback guard (origin allow-list, literal-loopback-only
+  Host gate, token check) that B3 (#61) MUST mount in front of its backend —
+  see `docs/security/desktop.md` for the threat model and the binding B3
+  integration contract.
+- Invariant: every new transport-facing seam threads the security barrel
+  (`desktop/main/security/index.ts`); the frozen B2 specs
+  (`src/__tests__/b2-*.test.ts`) plus the B1 secure-defaults spec are the
+  regression family that keeps the webPreferences, CSP, guard, and token
+  contracts from eroding.
 - The installer is unsigned at this stage; signing/updates are Workstream E5
-  (#88). Transport/CSP/loopback hardening is Workstream B2 (#60).
+  (#88).
 
 ## Reserved namespaces
 
--- a/desktop/main/index.ts
+++ b/desktop/main/index.ts
@@ -9,8 +9,15 @@
 import { existsSync } from 'node:fs';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
-import { app, BrowserWindow } from 'electron';
+import { app, BrowserWindow, ipcMain } from 'electron';
 import { registerAppProtocol, registerAppSchemePrivileges } from './protocol.js';
+import {
+  getLoopbackGuard,
+  getLaunchToken,
+  initializeLaunchToken,
+  initializeTransportSecurity,
+  resolveSecurityConfig,
+} from './security/index.js';
 
 const DEV_URL = 'http://localhost:5173';
 
@@ -23,6 +30,23 @@ function isDevArgv(): boolean {
 function devStartUrl(): string | undefined {
   const envUrl = process.env.ELECTRON_START_URL;
   return envUrl !== undefined && envUrl.length > 0 ? envUrl : isDevArgv() ? DEV_URL : undefined;
+}
+
+/**
+ * Navigation allow-list (issue #60): the renderer may only navigate itself to
+ * app:// targets; in dev mode the dev server origin is additionally allowed
+ * so vite's full-page reloads keep working. Everything else — remote http(s),
+ * file://, even loopback http in production — is denied.
+ */
+function isAllowedNavigationTarget(url: string): boolean {
+  if (url.startsWith('app://')) return true;
+  const dev = devStartUrl();
+  if (dev === undefined) return false;
+  try {
+    return new URL(url).origin === new URL(dev).origin;
+  } catch {
+    return false;
+  }
 }
 
 /** Directory of the compiled module (dist/main), valid under ESM. */
@@ -65,6 +89,18 @@ export function createMainWindow(): BrowserWindow {
     },
   });
   win.once('ready-to-show', () => win.show());
+  // Navigation lockdown (issue #60, AC4): the only renderer-initiated
+  // navigations are app:// targets (plus the dev server origin in dev mode);
+  // window.open is denied outright — no popups, no new windows, ever.
+  // NOTE: the window-open handler lives on webContents (the only surface that
+  // exists in Electron >= 44 typings AND at runtime — verified by probe:
+  // win.setWindowOpenHandler is undefined); will-navigate is webContents-native.
+  win.webContents.on('will-navigate', (event, url) => {
+    if (!isAllowedNavigationTarget(url)) {
+      event.preventDefault();
+    }
+  });
+  win.webContents.setWindowOpenHandler((): { action: 'deny' } => ({ action: 'deny' }));
   // Surface load failures instead of leaving a silent blank/hidden window
   // (e.g. dev server down, renderer resources incomplete).
   win.webContents.on('did-fail-load', (_event, errorCode, errorDescription, validatedURL, isMainFrame) => {
@@ -111,6 +147,24 @@ export function bootstrap(): void {
   registerSecondInstanceHandler();
 
   app.whenReady().then(() => {
+    // Transport security (issue #60, B2): resolve config, mint the per-launch
+    // token (main-process memory only), serve it to the renderer ONLY through
+    // this IPC handler, and construct the loopback gate that B3's backend
+    // host MUST mount in front of every route (docs/security/desktop.md).
+    const securityConfig = resolveSecurityConfig();
+    initializeLaunchToken();
+    ipcMain.handle('desktop:get-token', () => getLaunchToken());
+    initializeTransportSecurity({
+      token: getLaunchToken(),
+      tokenHeaderName: securityConfig.tokenHeaderName,
+      allowedOrigins: securityConfig.allowedOrigins,
+    });
+    if (getLoopbackGuard() === null) {
+      // Fail fast and audibly rather than booting an unguarded transport.
+      console.error('[trainingapp-desktop] transport security failed to initialize; refusing to start');
+      app.quit();
+      return;
+    }
     // Dev mode loads the vite dev server and needs no packaged renderer.
     if (devStartUrl() === undefined) {
       const root = resolveRendererRoot();
--- a/desktop/main/index.ts
+++ b/desktop/main/index.ts
@@ -165,6 +165,10 @@ export function bootstrap(): void {
       app.quit();
       return;
     }
+    // NOTE: in dev mode the app:// handler is intentionally NOT registered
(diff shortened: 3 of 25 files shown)


zaxbysauce:
## Swarm PR Review — #96 (Electron desktop transport-security hardening, issue #60)

**Method:** 5 parallel explorer lanes (correctness, security/trust-boundaries, dependency/deployment, docs/intent-vs-actual, tests/falsifiability) → independent reviewer (empirical: ran the full test suite, mutation-tested the guard, built the actual Vite output to check the CSP hash claim) → adversarial critic (challenged severities, checked fix-suggestions for new risk).

### PR intent
Workstream B2 builds the transport-security layer B1 (#95) deliberately deferred: per-launch token, origin/Host-gated loopback request guard, strict CSP + COOP/COEP/CORP on every `app://` response, navigation lockdown, and a token bridge exposed to the renderer via `contextBridge`. It also documents a binding integration contract for the not-yet-written B3 backend (#61).

### Implementation summary vs. claims
All major claims in the PR body were independently verified against the diff:
- `config.ts` defaults (`X-Desktop-Token`, `app://*`) — verified.
- `docs/security/desktop.md` threat model + B3 contract — present, matches described obligations.
- `getLoopbackGuard()` exported and importable — verified.
- Stale "deferred to B2" language in `desktop/README.md` — removed, replaced with real posture doc.
- `web_ui/` byte-zero diff — verified empty.
- No new dependencies; `crypto.randomBytes` is a Node builtin.
- tsconfig/CI wiring claims (new files compiled, new specs picked up by `npm --prefix desktop test`, preload CommonJS enforced by `verify-preload-format.mjs` in CI) — all verified against the actual workflow file.
- **Full test suite run: 79/79 tests pass, 12/12 files, no skips.** Mutation-testing (neutering the origin check, widening `LOOPBACK_HOSTS` to accept `localhost`) confirmed the guard's tests assert on real behavior, not stub echoes.

### Confirmed findings (all non-blocking; ship-safe today)

1. **[Minor] B3 integration contract has a gap that will bite the next PR, not this one.** `LoopbackGuardRequest` (`desktop/main/security/loopback-guard.ts:29-32`) carries no `method` field, so the guard cannot exempt CORS preflight `OPTIONS` requests, and its `fail()` responses never set `Access-Control-Allow-Origin`. Separately, the contract's own code snippet in `docs/security/desktop.md:87-90` (`const verdict = guard(req); if (verdict) return verdict;`) won't work against a raw Node request — `new URL(req.url)` throws on a relative path, and `req.headers` has no `.get()`. Both are the same root cause: the contract and interface were written ahead of any real consumer, and `getLoopbackGuard()` has exactly one caller today (`desktop/main/index.ts:162`, a null-check) — nothing in this PR's runtime path actually reaches the defect, and any failure mode fails closed (403/401/throw), never open. Recommend fixing before #61 (B3) is implemented against the current contract, since an implementer could otherwise "satisfy" the binding contract by exempting OPTIONS *before* the origin/host checks or setting `Access-Control-Allow-Origin: *` — that would be a real vulnerability, just not one this PR ships. Fix: add `method` to the interface, handle preflight *after* origin/host validation, and correct the doc's request-shape example.

2. **[Minor, latent only] CSP sha256 pin is tested against source `web_ui/index.html`, but the packaged app serves the Vite-built `dist/index.html`.** Independently rebuilt the PR's exact `index.html` with Vite 6 (default esbuild minify) — the built and source inline-script hashes are byte-identical today, so there is no live defect. But nothing pins the *built* artifact, so a future minify-config change would silently desync the CSP pin from what's actually served (failure mode: CSP blocks the theme-bootstrap script, not a security hole). Consider hashing `dist/index.html` in `desktop:build` or a post-build check.

3. **[Minor] One documented security invariant is enforced only by a script that doesn't run in CI.** `desktop/repro/check-b2.sh`'s `token_storage_guard()` (grep for `localStorage`/`sessionStorage`/URL-param token leakage) has no vitest equivalent — verified by grepping every file under `desktop/src/__tests__/` for those patterns (zero hits). The other 6 checks in that script are redundant with the vitest suite already wired into CI, but this one isn't. The property holds today (both greps clean against the PR tree), so this is a regression-protection gap, not an active issue. `docs/security/desktop.md` calls it "structurally enforced," which is currently inaccurate. Cheapest fix: add a one-line CI step running that specific check (the workflow already runs on `windows-latest` with Git Bash available); porting the two greps into vitest is a reasonable alternative but not required.

4. **[Informational] Misleading comment on a non-issue.** `loopback-guard.ts:111` comments the token equality check as "constant-time enough," but it's a plain `!==` string compare. Given the threat model (per-launch random 32-byte token, loopback-only, process-lifetime secret), timing-attack exploitability is negligible — anyone in a position to run a timing attack already has cheaper local-compromise paths. The actionable part is just the comment: fix it or use `crypto.timingSafeEqual`.

5. **[Note, no action required]** Dev-only origin allow-list uses `endsWith('*')` prefix matching, so a pattern like `http://localhost:5173*` would also match `http://localhost:5173.evil.com`. Gated behind `!app.isPackaged` and an explicit env opt-in, so not reachable in production; worth a doc line if anyone hand-writes a dev origin pattern.

### Pre-existing / out of scope
None flagged — this is a new subsystem (all files added or newly touched for B2).

### Disproved / downgraded claims
- Initial explorer-lane severities on the CORS-preflight finding (Major/High) and the check-b2.sh gap (High) were both downgraded by the reviewer+critic pass: nothing in this PR's shipped runtime path reaches either defect, both fail closed, and the "vulnerability" framing requires a not-yet-written consumer (B3) to implement the contract incorrectly.
- The claim that Vite minification could have already desynced the CSP pin was tested empirically and disproved for the current build config (byte-identical hashes).

### Test / coverage gaps
- AC3 ("token never in localStorage/sessionStorage/URL") is real and enforced today, but only by a script CI doesn't run (see finding #3).
- No other coverage gaps found; mutation testing confirmed the security-critical assertions (origin allow-list, Host gate, token equality, CSP headers on success/403/404, navigation lockdown, token non-persistence) all fail when the underlying logic is weakened.

### Verdict
**APPROVE-WITH-FOLLOWUP.** No finding is blocking or exploitable in code that ships in this PR; every failure mode fails closed. The one item worth prioritizing before #61 (B3) lands is amending `docs/security/desktop.md`'s integration contract and the `LoopbackGuardRequest` interface (finding #1) so the next implementer doesn't build against a contract whose own example is broken and whose interface can't express a safe CORS exemption.

---
*Reviewed via a 5-lane parallel explorer pass, independent opus-reviewer validation (including running the full test suite and an empirical Vite-build comparison), and an adversarial critic challenge on severity calls.*

Issue #86, opened 2026-09-05
[Workstream E] PR 3 of 6: Offline and low-RAM validation matrix on the reference laptop


zaxbysauce:
**Roadmap:** E3 · **Workstream E: Hardening, CI and release**
**Sequence:** PR 3 of 6 (tracking slot; dependencies define execution order)
**Roadmap baseline:** 2026-09-05 · **Priority:** P0 · **Repo revision:** b2ddda0

## Outcome
The full v3 app — Electron shell, chosen native backend (A7's ADR), bundled models, default packs — is run through a defined scenario matrix on the actual reference laptop (12th-gen i5, 16 GB RAM, Intel integrated graphics, Windows 11), with pass/fail and measured numbers recorded in `bench/RESULTS.md`. This is a PROVING gate: it does NOT fix defects it finds (file a new issue against the owning workstream and link it here) and does NOT re-run the CI-scale automated suite (E4) — it exercises real hardware conditions CI cannot simulate: actual RAM pressure, no network egress, cold boot, degraded cross-origin isolation.

## Evidence motivating this slot
This issue exists because the audit found the core "interactive speed on minimum hardware" claim unverified in every dimension it tests:
1. **No measured numbers exist anywhere.** `tests/test_rag_performance.py:25` and `tests/test_low_end_hardware.py:16,31` are module-level skipped. README/CHANGELOG carry only qualitative estimates ("~5-10 tok/s"; tiered "~5-7/~10-15/~15-20 tok/s" at [README.md:67,74,79](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/README.md#L67)).
2. **The desktop LLM loader deterministically refuses to load on a real 16 GB laptop today**: `llm_interface.py:400-412` requires free RAM >= 4x GGUF size (~12.4 GB free for the bundled ~3.1 GB model) while Windows 11 + browser + app typically leaves 8-11 GB free — the exact scenario A3 fixes and this matrix must confirm is fixed on real hardware, not the synthetic 4/16 GB numbers in `tests/test_llm_interface_memory_budget.py`.
[...]

Issue #67, opened 2026-09-05
[Workstream B] PR 9 of 9: Renderer integration — move Documents/ingest/settings to the API, Electron-aware mode, model/status UI


zaxbysauce:
**Roadmap:** B9 · **Workstream B: Electron shell and desktop backend**
**Sequence:** PR 9 of 9 (tracking slot; dependencies define execution order)
**Roadmap baseline:** 2026-09-05 · **Priority:** P0 · **Repo revision:** b2ddda0

## Outcome
Closing this makes every renderer surface touching documents, ingestion, or settings use `ApiClient` against the Electron-hosted backend when running inside Electron, while the pure-browser wllama/IndexedDB path keeps compiling as a secondary mode. Does NOT certify Knowledge Pack UI (C7/#74) or the Learn panel (D6/#82) — only that the core chat/documents/settings loop works end-to-end against the desktop backend with no reload across restarts.

## Evidence motivating this slot
- "The renderer needs no new client code" is FALSE and independently confirmed by the critic: [web_ui/src/pages/ChatPage.tsx:437-453](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/web_ui/src/pages/ChatPage.tsx#L437-L453) — CONFIRMED, this is the ONLY place with an api/browser mode branch: `if (mode === 'api') { ... streamManager.startSSEStream(url, {...}) } else { // Browser-local mode — RAG pipeline AsyncGenerator. }`.
- [web_ui/src/pages/DocumentsPage.tsx:1-18](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/web_ui/src/pages/DocumentsPage.tsx#L1-L18) — CONFIRMED: imports `extractDocument`/`TextChunker`/`document-store`/`profile`/`embedding-service`/`vector-index`/`keyword-index` unconditionally — no `mode==='api'` branch anywhere. `critic_sol_verdict.md` risk 2: "Documents remains IndexedDB/EdgeVec/FlexSearch... Documents/ingest/packs UI must be built against the API."
[...]

Issue #65, opened 2026-09-05
[Workstream B] PR 7 of 9: Hybrid retrieval (sqlite-vec + FTS5 -> RRF -> reranker) with eval parity


zaxbysauce:
**Roadmap:** B7 · **Workstream B: Electron shell and desktop backend**
**Sequence:** PR 7 of 9 (tracking slot; dependencies define execution order)
**Roadmap baseline:** 2026-09-05 · **Priority:** P0 · **Repo revision:** b2ddda0

## Outcome
Closing this gives the Electron-hosted backend a hybrid vector+FTS5 retrieval pipeline (RRF fusion, worker-thread reranking, a re-calibrated relevance floor) feeding `/search` and `/ask`/`/ask/stream`, reaching recall/MRR parity with the web_ui browser-mode baseline on A4's eval set. Does NOT certify Knowledge Pack recency weighting (hook reserved here, implemented in C4/#71) — only core retrieval quality/latency.

## Evidence motivating this slot
- `audit_report.md` gap #10 cites `rag-orchestrator.ts:148-163` — "ettin's score histogram 'is not yet re-validated'; floor remains `0.2`" (`MIN_CROSS_SCORE=0.2`) — CONFIRMED. This issue's re-calibration must run against whatever reranker A5/#55 actually selects, not assume ettin's 0.2 floor transfers.
- [web_ui/src/lib/search/reranker.ts:28-38](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/web_ui/src/lib/search/reranker.ts#L28-L38) — CONFIRMED: `MAX_CHUNK_COUNT=500`, `RERANK_INPUT_CAP` of 50, comment "~11 sequential WASM batches of a ~32M-param ModernBERT reranker (2-5s on the target i5)." The native worker-thread path should be faster per `implementation_plan.md:40`'s (unbenchmarked) "2-4x faster than WASM" claim — this issue must MEASURE, not assume, before setting a native candidate cap.
[...]

Issue #64, opened 2026-09-05
[Workstream B] PR 6 of 9: Profile-scoped ingestion, migrations, backup and recovery on the SQLite store


zaxbysauce:
**Roadmap:** B6 · **Workstream B: Electron shell and desktop backend**
**Sequence:** PR 6 of 9 (tracking slot; dependencies define execution order)
**Roadmap baseline:** 2026-09-05 · **Priority:** P1 · **Repo revision:** b2ddda0

## Outcome
Closing this gives a working ingest pipeline (extract -> chunk -> embed -> write) against B5's schema, a defined profile ownership model (default OS-user-scoped single profile, optional named profiles), and backup/restore plus corruption recovery. Does NOT certify hybrid retrieval quality (B7/#65) — only that documents get in, stay deduplicated across re-ingestion, and survive restarts/corruption.

## Evidence motivating this slot
- [web_ui/src/lib/storage/profile.ts:60-76](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/web_ui/src/lib/storage/profile.ts#L60-L76) — CONFIRMED: `getProfilePrefix()` reads/mints a single random localStorage id via `mintPrefix()`; no UI or API to view/switch/name profiles. `critic_sol_verdict.md` risk 9: "'One file per profile' has no ownership, switching or migration design." This issue must make an explicit documented choice rather than porting the "one random id" pattern into Electron.
- Stale-chunk retention on re-ingest is a confirmed desktop defect this pipeline must not reproduce: `audit_report.md` gap #8 cites `document_processor.py:349` (path-based doc id), `vector_store.py:604,650-663` (chunk-id existence check skips colliding ids without deleting), `rag_engine.py:298-303` (never deletes the old document first) — CONFIRMED per the audit's citations. B5's content-hash chunk ids are the fix; this issue's pipeline must delete-before-reingest at the `doc_id` level.
- `implementation_plan.md:59` ("installation is a copy + attach, no re-embedding") and the existing Clear-Cache precedent (`SettingsPage.tsx:724-726` `deleteEdgeVecBlob`, per `critic_sol_verdict.md` finding 11 REFUTED as already-correct) set the bar this issue's simpler SQLite-mode Clear Cache (delete the one profile DB file) must not regress below.

## Prerequisites and ownership
[...]

PR #97, opened 2026-09-08
feat(desktop): Electron backend host answering the frozen API contract (issue #61)

### LINKED ISSUE #61: [Workstream B] PR 3 of 9: Backend host in Electron implementing the frozen API contract, per ADR-0003

zaxbysauce:
Closes #61 (Workstream B PR 3 of 9: Backend host in Electron implementing the frozen API contract, per ADR-0003)

## Root Cause

Not a defect: issue #61 is a roadmap implementation slot. At base a00f8d1 the Electron shell (B1 #59 + B2 #60) stopped at transport security — `desktop/main/backend/` did not exist, `desktop/main/` contained no HTTP listener (`grep "createServer|listen("` → zero hits), the B2 loopback guard's only consumer was the fail-fast null-check at `desktop/main/index.ts:162`, and `contracts/tests/run_conformance.py:12-13` reserved `--base-url` mode for "the Electron host from WS-B #61 later" with nothing to aim it at (any `--base-url` run failed with `httpx.ConnectError`). Full gap analysis in the trace: `.agents/issue-traces/61-electron-backend-host-frozen-api-contract/04-root-cause.md`.

## Fix

One guarded loopback listener, two backends behind it, one selector:

- `desktop/main/backend/server.ts` — node:http listener with a SINGLE guard call site before routing, OPTIONS short-circuit, and sidecar proxy; the request adapter builds the absolute URL the guard requires and folds header case through `Headers`. Serves all 14 operations of `contracts/api.openapi.yaml` (404 unknown / 405 wrong-method); SSE emits `data: {json}\r\n\r\n` frames payload-discriminated as token/done/error with exactly one terminal (client disconnect → `done{cancelled:true, sources, context_length}`); JSON/multipart body caps (413); contract-shaped 422 validation; CORS preflight answered only AFTER the guard's gates.
- `desktop/main/backend/engine.ts` + `types.ts` — `StubEngine`: documented, contract-shaped stubs for every route (real inference lands with B4 #62, ingestion B6 #64, retrieval B7 #65; stubs honestly return named not-implemented messages, never fake success). Settings mirror `api_server.py` (12 keys, `rag_*` bounds, `rag_chunk_overlap >= rag_chunk_size → 400`). `RESERVED_PROFILE_HEADER_NAME = 'X-Profile-Id'` is accepted on every route and ignored until B6 (#64), so profile-scoped storage cannot change the wire contract later.
- `desktop/main/backend/sidecar-manager.ts` — SidecarManager (per ADR-0003's sidecar option): spawn with loopback bind config (`API_HOST=127.0.0.1`, `API_PORT=<port>`; `--host/--port` args for uvicorn-style launchers), bounded `GET /health` readiness with retry/backoff (60 s default), bounded-retry restart on unexpected exit with exponential backoff and a loud give-up, graceful-then-kill shutdown (SIGTERM → grace → SIGKILL; on Windows both map to TerminateProcess — documented), child-pipe draining (an undrained pipe blocks a chatty child mid-boot), and timer cleanup so app quit leaves no orphan.
- `desktop/main/backend/index.ts` — THE import seam for B4-B9: `createBackendHost(config)` returns one of two implementations of the same `BackendHost` interface (`start() → {mode, port, url}`, idempotent `stop()`), chosen only by `backend.mode` (env `TRAININGAPP_DESKTOP_BACKEND_MODE`, default `"node"` while ADR-0003 #57 is open — recorded assumption; flipping is one line). Sidecar mode = the SAME guarded listener fronting a loopback proxy to the child: strips `X-Desktop-Token` before forwarding, forwards `X-Profile-Id` unchanged, tears down half-open streams.
- `desktop/main/backend/dev-server.ts` — headless compiled entry (`--port-file/--mode/--token/--sidecar-*`), electron-free, exits when stdin closes; what CI and the acceptance checks drive.
- `desktop/main/security/defaults.ts` (new) — `DEFAULT_TOKEN_HEADER_NAME` / `DEFAULT_ALLOWED_ORIGINS` moved to an electron-free module (the npm `electron` package is a CJS string export; a named `import { app }` breaks plain-node boot); `config.ts` re-exports for import compatibility; `loopback-guard.ts` imports the defaults — rule set byte-identical, B2 frozen families untouched.
- `desktop/main/index.ts` + `desktop/preload/index.ts` — bootstrap starts the host behind the guard, registers `desktop:get-backend` IPC (`desktopApi.getBackendInfo()` → `{mode, port, url}` for B9; never web storage, never a URL), and stops the host on `will-quit`.
- `desktop/scripts/run-conformance-host.mjs` + CI job `backend-conformance` (in `.github/workflows/desktop-build.yml`) — compiles the host, boots it headless with a random run token, measures cold start to first 200 `GET /health`, and runs the frozen suite via a HARNESS-ONLY loopback proxy that injects the token so every request still traverses the real guard (never bypassed, never shipped). The suite runs via ASYNC spawn — a synchronous spawn blocks the harness event loop and stalls the proxy.

- Defect class: a new server/transport surface that bypasses the security barrel — mounts no guard, hard-codes the token header name, imports security config through a semantics-changing path, or re-introduces an electron import into a headless module.
- Sweep result: 6 predicates, 10 hits, all dispositioned (1 planned listener site, 1 canonical token definition, 1 compliant guard site, 1 documented config.ts electron exception; CI job + IPC pairing hits are the demanded invariants themselves) — `.agents/issue-traces/61-electron-backend-host-frozen-api-contract/08a-recurrence-sweep.md`.
- Guardrail: runtime trust-boundary assertion — the guard-spy spec in `desktop/src/__tests__/b3-server.test.ts` proves every request (mixed 200/401/403/404 classes) traverses the guard exactly once before any handler and that an unauthorized `/ask` never reaches the engine. Demonstrated to bite: discarding the guard verdict turns 4 of 9 b3-server tests RED (reviewer-run mutation probe); neutering the will-quit handler turns the bootstrap-wiring spec RED (final-critic mutation probe). Backed by CI check C6 probing every frozen route through the real guarded listener.

## Tests

- Regression test: `npm --prefix desktop test` → `Test Files 21 passed (21)` / `Tests 125 passed (125)` (15 pre-existing B1/B2 files unchanged and green + 6 new b3 files / 31 tests)
- Impacted suite: `python contracts/tests/run_conformance.py --asgi api_server:app` → `CONFORMANCE: PASS 12/12 checks passed, 0 skipped` (Python side untouched)
- New CI gate: `node desktop/scripts/run-conformance-host.mjs` → `CONFORMANCE: PASS 10/10 checks passed, 2 skipped` (skips are --asgi-only checks by suite design) against the Electron-hosted backend, cold start 126-135 ms vs the 15000 ms bound
- Lint/type/build/security checks: `npm --prefix desktop run compile` → 0 TS errors; `git diff a00f8d1..HEAD -- web_ui/ contracts/ api_server.py` → EMPTY (ChatPage.tsx byte-unchanged per AC2's zero-renderer-changes constraint)
- Acceptance checks C1-C10: all RED/ERROR at base → GREEN at head via `repro-check.sh run` (logs in the trace `repro/` dir); 5 CHECK_WRONG AMEND rows with fresh replays documented (driver-level bugs found during the flip, each fixed and re-frozen)
- Deferred-work scan: `scan-deferred.sh a00f8d1` → clean
- Sidecar-mode E2E (beyond the gate): the real selector in sidecar mode with `python -m uvicorn api_server:app` as the child and a real GGUF (RAG_GGUF_PATH) → conformance through the guarded proxy 9/10 PASS; the one FAIL (`ask_stream` tokens=0) is the real engine's streaming behavior for the LFM2.5-VL mmproj model on this box (proxy faithfully delivered 200 + text/event-stream + exactly 1 terminal, 0 malformed) — real-model streaming is B4 #62 scope

## Regression Protection

- `desktop/src/__tests__/b3-backend-default.test.ts` — selector contract (default→node, explicit modes, env flip, loud failure without sidecar.command, X-Profile-Id constant)
- `desktop/src/__tests__/b3-backend-selector.test.ts` — two different implementations, same duck-typed interface, config-only swap
- `desktop/src/__tests__/b3-sidecar-manager.test.ts` — loopback bind args/env, bounded health retry/backoff, graceful-then-kill ordering, idempotent stop
- `desktop/src/__tests__/b3-sidecar-lifecycle.test.ts` — bounded restart with observed backoff growth, give-up after bound (no tight loop), no orphan after stop
- `desktop/src/__tests__/b3-bootstrap-wiring.test.ts` — the REAL runtime path: bootstrap() starts the host, desktop:get-backend returns the live handle, guarded /health 200 with token / 401 without, and will-quit closes the port (net-probe asserted; mutation-proven to bite)
- `desktop/src/__tests__/b3-server.test.ts` — guard-spy (every request crosses the gate exactly once before any handler), SSE disconnect → single cancelled-done terminal, proxy strips X-Desktop-Token / forwards X-Profile-Id, OPTIONS only after gates, settings cross-field rule, 422/405/auth surfaces
- Boundary/adversarial: oversized bodies (413), malformed JSON (422), whitespace question (422), unknown route (404), wrong method (405), SSE double-terminal prevention, half-open proxy teardown
- Test drift review: 15 pre-existing desktop spec files and the Python test surface untouched and re-ran green; contracts/api.openapi.yaml and run_conformance.py byte-unchanged

| Acceptance criterion (from issue #61) | Evidence (command + output, or test name) |
|---|---|
| Conformance green: contracts/tests pass against the Electron-hosted backend in CI | CI job `backend-conformance` (.github/workflows/desktop-build.yml) → `node desktop/scripts/run-conformance-host.mjs` → `CONFORMANCE: PASS 10/10 checks passed, 2 skipped` (reproduced locally 3x: 10/10); C1 acceptance check GREEN |
| Renderer connects: ChatPage mode==='api' streams through /ask/stream with zero ChatPage.tsx changes | `git diff a00f8d1..HEAD -- web_ui/` EMPTY; C2 wire-shape check GREEN (the exact request/SSE shape `web_ui/src/lib/api/streaming.ts` + `ChatPage.tsx:437-457` consume); manual procedure + port discovery (`desktopApi.getBackendInfo()`) documented — B9 completes the UI wiring |
| Health check: GET /health returns 200 within N seconds cold (N recorded) | bench/RESULTS.md "Desktop backend host cold start (issue #61)": 135 ms / 132 ms (devstation, bound 15000 ms); harness prints the measurement every CI run (independent runs: 131/126/135 ms) |
| Sidecar lifecycle: external kill → bounded-retry restart; app quit → no orphan | b3-sidecar-lifecycle + b3-sidecar-manager specs GREEN (injectable spawn; will-quit stop mutation-proven); real-uvicorn sidecar E2E executed; on-hardware PyInstaller runs tracked for E3 #86 |
| Backend-mode selector: flipping backend.mode swaps implementation without touching B4-B9 call sites | b3-backend-selector + b3-backend-default specs GREEN (duck-typed interface, config-only swap, env flip); C5/C8 GREEN |
| Required scope: server registers EVERY contract route behind B2's guard middleware | C6 GREEN: all 14 operations 401 no-token / 401 wrong-token / answered with valid token via proxy, unknown path 404; single guard call site in server.ts |
| Required scope: sidecar manager spawn/health/backoff/restart/shutdown | b3-sidecar-manager GREEN; real uvicorn E2E through the guarded proxy 9/10 |
| Required scope: backend.mode selector so B4-B9 import one interface | desktop/main/backend/index.ts (`createBackendHost`/`resolveBackendMode`); b3-backend-default GREEN |
| Required scope: contracts/tests in CI against this host | `backend-conformance` job step "Run API contract conformance against the Electron-hosted backend" |
| Required scope: docs/security/desktop.md host:port binding + spawn args/working directory | C9 GREEN — "B3 backend host (issue #61, implemented)" section (random free port, guard mounting, `backend.mode`, sidecar spawn args incl. `--host`, working directory, harness proxy, X-Profile-Id, renderer discovery) |
| Evidence note: reserve a profile-identifying header/param slot for B6 | `RESERVED_PROFILE_HEADER_NAME = 'X-Profile-Id'` (types.ts); accepted on all routes, forwarded by the sidecar proxy; C10 GREEN |

## Invariant Audit

The governing invariant doc for this surface is `docs/security/desktop.md` ("Architecture invariants" + binding "B3 integration contract"):

- Renderer is untrusted input: not touched — window/security flags unchanged; b2 secure-defaults family green.
- Every transport seam threads the security barrel (`desktop/main/security/`): touched AND extended — the new seam composes `getLoopbackGuard()` via a single call site (guard-spy spec + mutation probe); token header name read from config, never hard-coded (sweep predicate 2).
- Loopback-only transport: touched — binds `127.0.0.1` on an OS-assigned random free port; R1/R2/R3 guard rules byte-unchanged (loopback-guard.ts only re-homed its constant imports); B2 regression family (b2-loopback-guard/origin/security-config/token-*) all green.
- B3 contract items 1-5: all satisfied and independently verified (implementation review + final critic, line-level) — random free port, guard before EVERY route incl. OPTIONS-after-gates, absolute-URL adapter, configured token header name, no token/URL logging at info level.
- Other repo invariants (B1 window flags, protocol handler containment, CSP pin): not touched — `b2-csp-pin` and B1 families green; protocol.ts untouched.

## Risk and Rollback

- Risk level: low-medium — new code is additive behind one import seam; the only pre-existing files touched are `desktop/main/index.ts` (bootstrap wiring, fail-loud on host-start failure), `desktop/preload/index.ts` (additive bridge method), `security/config.ts`+`loopback-guard.ts` (constant re-home with re-export, import-compatible), CI workflow (new job), and docs.
- Rollback: revert the two commits on `feat/ws-b3-electron-backend-host-issue-61` (33dc557, 714bbbf); no migrations, no data, no dependency changes; the Python server and web_ui are untouched.
- Residual risk: the stub engine returns honest not-implemented responses for ingestion and template answers for /ask until B4/B6/B7 land (documented in code, README, docs, CHANGELOG); reference-laptop PyInstaller kill/restart and packaged-app manual streaming remain E3 #86 hardware validations (documented assumption A5).

## Waivers (or none)

none

## Merge status

PR head: 714bbbf6edc012736e618c6f9060e8c934c57ab5

## Issue Closure

Closes #61

## Full-Resolution Contract note (issue-tracer trace)

Trace: `.agents/issue-traces/61-electron-backend-host-frozen-api-contract/` — plan critic APPROVE (Round 2, 7 revisions incorporated), 10/10 acceptance checks RED→GREEN with a frozen red checkpoint (5 disclosed CHECK_WRONG driver amendments with fresh replays), recurrence sweep with a mutation-proven guardrail, independent implementation review APPROVE (33dc557 + delta 714bbbf), final critic APPROVE (Round 2). Independent verifier identities are recorded in the trace `state.md` Gates table. No waivers.

<a href="https://cubic.dev/pr/ZaxbyHub/trainingapp/pull/97?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

## Review round 2 — feedback resolution (3e98c0b)

A full swarm review (6 explorer lanes + 11 risk-family micro-lanes + independent reviewer validation) plus the PR comment reviews produced a validated findings set: PRR-001..013 from the review handoff and the external comment-review findings (#1-#11, suspicious items, test gaps). All are resolved in commit 3e98c0b (16 files, +337/-74); full disposition ledger: .zcode/pr-review/pr97-714bbbf/closure-ledger.md. Highlights:

- **Security/correctness:** settings validation own-property guard (prototype-named keys now 422); failed sidecar start() can no longer arm an orphaned restart chain (full stop + starting flag + regression spec); oversized bodies now get an observable 413 instead of a connection reset; will-quit holds the quit until backend shutdown completes; conformance harness failure branch exits non-zero (was printing FAILED yet exiting 0); harness error path exits immediately instead of hanging the runner.
- **Lifecycle:** restart budget is a rolling window (restartWindowMs, default 60s) matching the docs; sidecar.env can no longer override the loopback bind config; partial-init cleanup in both host start() paths; give-up is surfaced fail-loud via config.onGiveUp (dialog in the Electron entry), not console-only.
- **CI/coverage:** workflow now triggers on contracts/**; duplicated vitest step dropped from backend-conformance (desktop-build still runs it); 15-minute job backstop; 5 new regression tests (130 total, 22 files).
- **Docs:** bind-close race named, sidecar bind-by-contract risk stated, harness --token argv tradeoff documented (dev/CI-only), guard-mounting wording corrected, X-Profile-Id validation deferred to B6, bench/RESULTS.md provenance channel registered.

Closed-as-reviewed (not defects of this diff): web_ui client-type divergences (renderer wiring is B9 #67 — the server conforms to the frozen YAML, conformance PASS); 60MB multipart wire cap vs 50MB file cap (deliberate encoding-overhead headroom, documented at server.ts:25); readBody abort-path 413 (writes only to an already-dead socket). Non-blocking follow-ups: README preload wording predates getBackendInfo; onGiveUp dialog path verified by reading (no dedicated spec).

Gates at 3e98c0b: compile green; vitest 130/130 (22 files); frozen conformance suite CONFORMANCE PASS 10/10 against the modified host; independent reviewer APPROVE (20/20 after one revision); final critic APPROVE (zero contract drift; re-ran the suite independently).


zaxbysauce:
## swarm-pr-review: PR #97 (Workstream B3 — Electron backend host, issue #61)

Automated multi-lane adversarial review (6 parallel explorer lanes: correctness, security, deployment safety, docs/intent-vs-actual, tests, performance/architecture), followed by independent re-verification of the highest-severity claims against source. Scope: 23 files, 2544 insertions — new `desktop/main/backend/{server,engine,sidecar-manager,index,types,dev-server}.ts`, `desktop/scripts/run-conformance-host.mjs`, `.github/workflows/desktop-build.yml`, 6 new test files, docs.

### PR intent
Wire a loopback-only HTTP backend host behind the existing security guard, serving all 14 contract operations via a `StubEngine` (honest not-implemented responses until B4–B9 land real inference/ingestion/retrieval), with a `SidecarManager` process-lifecycle option and a conformance-CI job to certify wiring + contract shape. This is explicitly scoped as WIRING/CONTRACT CONFORMANCE only, not real functionality.

### Implementation summary vs. claims
The PR body's specific, falsifiable claims were checked against code and are **mostly accurate**: guard-before-CORS-preflight ordering is correct, SSE framing (`data: {json}\r\n\r\n`, single-terminal invariant, `cancelled:true` on disconnect) is correctly implemented and well-tested, the 14-operation route table matches `contracts/api.openapi.yaml`, the 12 `rag_*` settings keys and `overlap >= size → 400` cross-field rule are correct, CORS reflects only allowlisted origins (no wildcard+credentials misconfig), and the loopback guard is called once, before every route including OPTIONS and SSE, with no bypass path found. `bench/RESULTS.md` and `CHANGELOG.md` entries check out.

Where the review found real gaps, they cluster in exactly the two places you'd expect for infra of this kind: **the safety net that's supposed to catch regressions (CI conformance gating)**, and **the not-yet-exercised failure paths in the sidecar lifecycle**.

---

### Confirmed findings — recommend fixing before merge

**1. `PUT /settings` validation bypass / prototype pollution** — `desktop/main/backend/engine.ts:126-161`
`SETTING_BOUNDS` is a plain object literal; the lookup `(SETTING_BOUNDS as Record<...>)?.[key]` is not own-property-guarded. `PUT /settings {"toString": 1}` (or `{"constructor": 1}`) resolves `bounds` to `Object.prototype.toString`/`.constructor` — truthy, so the "unknown setting" check at line 127 is skipped. Every subsequent bound check (`bounds.type`, `.min`, `.max`) reads `undefined` off that function object and falls through silently, so `applied.toString = 1` gets `Object.assign`'d onto `this.settings` at line 161. Result: a request that should 422 as an unknown key instead returns 200 and corrupts internal state. Independently re-verified by direct read (traced the exact `bounds` resolution). Fix: guard with `Object.prototype.hasOwnProperty.call(SETTING_BOUNDS, key)` or use a `Map`/`Object.create(null)`.

**2. Sidecar restart produces an orphaned background respawn loop, contradicting the PR's own "no orphan" claim** — `desktop/main/backend/sidecar-manager.ts:90-98,138-159`, `desktop/main/backend/index.ts:116-117`
When `waitUntilHealthy()` fails during `start()`, `stopChildOnly()` runs but `stopping` is never set to `true`, and `start()` throws before `SidecarBackendHost.start()` (index.ts:117) can assign `this.manager`. The child's `'exit'` handler still sees `stopping === false` and calls `scheduleRestart()` — but the manager instance is now unreachable from `stop()`, so nothing can cancel the pending restart timer. Up to `maxRestarts` further children can spawn in the background after the app has already concluded startup failed. Independently reached by two separate review lanes reading the same code path. Directly contradicts `sidecar-manager.ts:14-15`'s "leaves NO ORPHAN child" doc comment.

**3. Conformance CI job can report green on a genuinely failed run** — `desktop/scripts/run-conformance-host.mjs:184-199`
If the python suite exits `0` without emitting the `CONFORMANCE: PASS` marker in stdout, the script's own logic prints "FAILED" — then calls `process.exit(suiteCode)`, i.e. `exit(0)`. `.github/workflows/desktop-build.yml:97` reports this as a passing job. This defeats the entire purpose of the `backend-conformance` job this PR introduces. Independently confirmed by two lanes.

**4. Conformance CI job is never triggered by contract-only changes** — `.github/workflows/desktop-build.yml:6-13`
The `paths:` filter covers `desktop/**` and `web_ui/**` but not `contracts/**` — even though the job exists specifically to guard the frozen API contract, and the workflow's own pip cache key (line 78) already hashes `contracts/tests/run_conformance.py` as an input. A contract-only PR never runs this gate.

**5. Unhandled error paths hang the CI runner instead of failing fast** — `desktop/scripts/run-conformance-host.mjs:206-209`
`main().catch(...)` sets `process.exitCode = 1` but never calls `process.exit()`; the proxy's open listen handle and the piped host child keep the event loop alive, so a thrown error (missing python binary, cold-start timeout) leaves the job hanging until the runner's global timeout — no `timeout-minutes` is set anywhere in the workflow.

### Confirmed, lower severity — recommend fixing, non-blocking

**6. Multipart body cap enforces 60MB while the spec and error message say 50MB** — `server.ts:25` (`MULTIPART_BODY_CAP_BYTES = 60 * 1024 * 1024`, comment says "contract cap is 50 MB") vs `server.ts:438`'s own 413 message and `contracts/api.openapi.yaml`'s `/ingest/file` limit. A 51–59MB upload is silently accepted when the spec says reject.

**7. `POST /ingest/batch` returns a fake "successful empty batch" instead of an honest not-implemented stub** — `server.ts:453` hardcodes `engine.ingestBatch(0)` regardless of how many files were actually posted; `engine.ts:194-201` returns `{total_files:0, successful:0, failed:0, results:[]}` with 200 OK and no message — unlike the sibling `ingestFile`/`ingestDirectory` stubs, which honestly return `success:false` with an owning-issue message. A client posting 10 files gets a response indistinguishable from "nothing to do."

**8. `413` (body-too-large) responses never reach the client** — `server.ts:99-106`: on exceeding the cap, `readBody` calls `req.destroy()` then resolves `null`; the subsequent `sendJson(res, 413, ...)` call writes to an already-destroyed socket. Not a memory-safety issue (the cap itself is enforced incrementally, before full buffering), but the documented 413 contract response is unobservable — clients see a connection reset instead.

**9. `will-quit` does not await backend shutdown** — `desktop/main/index.ts:190-192` (compounds finding #2): the handler doesn't await `backendHost?.stop()`, so Electron can exit while sidecar SIGTERM/grace/SIGKILL cleanup is mid-flight, orphaning the child. `dev-server.ts:83` shows the correct pattern (`await host.stop().finally(...)`) for comparison.

**10. Restart counter is a lifetime cap, not the rolling window the option doc implies** — `sidecar-manager.ts:60,152` — `restarts` is incremented on every restart and never reset after a healthy run, so a sidecar that flaps 5 times total over days of otherwise-healthy operation permanently exhausts `maxRestarts` and gives up on the next unrelated crash.

**11. "Loud give-up" is just `console.error`; the host keeps accepting/proxying to a dead sidecar** — `sidecar-manager.ts:148`, `index.ts:113-115` — once restarts are exhausted, nothing marks the backend handle unhealthy; requests keep getting proxied to a port nothing is listening on. In a packaged Electron app, `console.error` on the main process is invisible to the user.

### Suspicious / worth a follow-up (not blocking, flagged for awareness)

- **`sidecar-manager.ts:118-125`** — caller-supplied `sidecar.env` is spread *after* the hard-coded `API_HOST=127.0.0.1`/`API_PORT` defaults, so it can silently override them. Not exploitable today (nothing in this PR feeds untrusted input into `sidecar.env`), but a latent foot-gun the moment a settings UI or CLI flag wires into it — recommend swapping merge order or stripping those two keys from caller env.
- **`server.ts` `proxyToUpstream`/`runAskStream`** — no `'error'` listener on the client `res` object (unlike the three other streams in the same function), a theoretical unhandled-EventEmitter-error crash path; and the routing dispatch (`Headers` construction, the `guard()` call) runs outside the request handler's `try/catch`, which only wraps the async IIFE further down.
- **`X-Profile-Id`** header is forwarded to the sidecar unfiltered/unvalidated (size/charset) even though currently inert — worth bounding before B6 gives it meaning.
- **Partial-init resource leak**: if `start()` throws between server creation and handle assignment in `NodeBackendHost`/`SidecarBackendHost`, a retry can orphan the first server (masked today by app-quit-on-error behavior).
- Full request body is still buffered into memory for `/ingest/file` and `/ingest/batch` even though it's discarded by the stub — expected to change with B6, flagged so it isn't forgotten.
- Docs: `sidecar-manager.ts:131-134`'s comment ("forward to the manager's own console at debug volume") doesn't match the code (child stdout/stderr silently drained, no logging at all) — safer behavior, misleading comment.

### Test / coverage gaps
The 6 new test files are, on net, **above average** for this class of PR — the SSE single-terminal invariant, guard-call-count, and sidecar restart-count/backoff-growth assertions are genuinely falsifiable, not theater, and one test file's comment documents a previously-fixed test-theater bug (an unawaited promise in a polling loop). Gaps:
- No test exercises the exact numeric boundaries the validation code enforces (413 cap bytes, question/query length limits, `n_results`/history count caps) — only obviously-valid/invalid cases.
- `b3-sidecar-manager.test.ts` uses real timers with tight windows (e.g. 30ms grace) rather than fake timers — a CI-flake risk.
- Sidecar mode's `start()` success path (spawn → healthy) isn't exercised in the selector/default tests, only the "missing command" rejection.
- One test hand-copies a validation error string rather than importing the source constant (fixture/production drift risk).

### Minor/low severity (optional cleanup)
405 responses omit the `Allow` header; `OPTIONS` on an unknown path returns 204 rather than 404; non-streaming `/ask` doesn't pass a `cancellationEvent` so client aborts don't stop in-flight stub work; CI workflow follows the repo's existing (pre-existing, not a regression) convention of floating-tag actions and no `permissions:`/`timeout-minutes:` blocks.

### Disproved / non-issues
No security vulnerability was found in the loopback guard, CORS handling, or the trust-boundary lane generally — the guard-call-site placement, origin-reflection gating, and bind-address behavior all match the documented threat model. The token-comparison-is-not-constant-time observation is an explicitly accepted, reasonable tradeoff given the loopback-only, per-launch-rotated design. All specific PR-body claims about SSE framing, contract operation count, settings key count, and CORS ordering were verified accurate.

### Verdict & merge recommendation
**Request changes.** The core wiring and security posture are sound, and most of the PR's own claims hold up under adversarial review — but three of the confirmed findings (settings validation bypass, sidecar orphan-restart loop, conformance-CI false-green/never-triggered) undermine exactly the guarantees this PR exists to establish for the workstream (contract conformance certification and safe process lifecycle), and every subsequent B4–B9 PR will build on this foundation. Recommend fixing findings **#1–#5** before merge; **#6–#11** and the test gaps are reasonable to batch into a fast follow-up PR if the team prefers to unblock B4 first, but should not be silently dropped.

---

zaxbysauce:
Full disposition of this review is now on the PR head (3e98c0b, 16 files +337/-74). Per-finding ledger with evidence: resolved in the PR description (Review round 2 section); the machine-readable version lives at .zcode/pr-review/pr97-714bbbf/closure-ledger.md in the workspace.

Summary against the numbered findings here:

- **#1 settings validation bypass** — FIXED: own-property guard on SETTING_BOUNDS (engine.ts); toString/constructor/hasOwnProperty now 422 as unknown setting. Regression test added (would have failed before).
- **#2 failed-start restart orphan** — FIXED: failed start() now runs full stop() AND a starting flag suppresses restart scheduling for exits during the health wait (a restart timer armed mid-boot could otherwise fire before the failure cleanup). Regression test pins exactly 1 spawn after a failed start.
- **#3 conformance false-green** — FIXED: failure branch exits non-zero when the sentinel is missing.
- **#4 contracts/** path filter** — FIXED: added to push + pull_request triggers.
- **#5 harness error path hangs runner** — FIXED: main().catch now process.exit(1); job got a 15-minute timeout backstop.
- **#6 60MB vs 50MB** — closed as deliberate: the contract caps the FILE (yaml: "max 50MB"); the 60MB wire cap covers multipart encoding overhead (documented at server.ts:25); the route is a stub that ignores content.
- **#7 /ingest/batch honesty** — documented+bounded: frozen BatchIngestResponse has no not-implemented slot; real multipart parsing/honesty lands with B6 (#64); STUB comments added.
- **#8 413 unobservable** — FIXED: readBody no longer destroys the request on cap-exceeded; the 413 flushes and Node closes the connection after. Regression test posts 1MB+1 and asserts a fetch-visible 413.
- **#9 will-quit not awaiting stop** — FIXED: preventDefault + await backendHost.stop() + app.quit() re-entry via a quitting flag.
- **#10 lifetime restart counter** — FIXED: rolling window (restartWindowMs, default 60s); backoff semantics unchanged (existing specs still pass).
- **#11 give-up surfacing** — FIXED: new config.onGiveUp seam; the Electron bootstrap wires it to a fail-loud error dialog (console.error alone was invisible in a packaged app). Clients keep getting contract-safe 502s.
- **Suspicious items**: env-override foot-gun fixed (loopback keys win, tested); res.error-listener theory disproved by an executable reproducer (writeHead on destroyed-not-sent responses does not crash node:http — .zcode/pr-review/pr97-714bbbf/f4-proxy-settlement.mjs); X-Profile-Id validation deferred to B6 (documented); partial-init leaks fixed in both hosts; drain comment corrected.
- **Test gaps**: boundary tests added (oversized 413, 2001-char question, n_results 11/0, prototype keys) + a new preload-bridge spec; suite is 130/130 across 22 files.

Validation at head: compile green, vitest 130/130, frozen conformance suite CONFORMANCE PASS 10/10 against the modified host; independent reviewer APPROVE (20/20) and final critic APPROVE (zero frozen-contract drift).

Issue #69, opened 2026-09-05
[Workstream C] PR 2 of 9: Python PackManager lifecycle (install/supersede/rollback/remove) with content-hash chunk ids and update-in-place


zaxbysauce:
**Roadmap:** C2 · **Workstream C: Knowledge packs and freshness**
**Sequence:** PR 2 of 9 (tracking slot; dependencies define execution order)
**Roadmap baseline:** 2026-09-05 · **Priority:** P0 · **Repo revision:** b2ddda06e86b63f81bdbedce220a3a2357bc9852

## Outcome
When this closes, the Python backend can install a pack.json-conformant Knowledge Pack (C1) into ChromaDB with correct content-hash chunk identity, and re-installing an updated version of the same pack replaces stale chunks instead of accumulating duplicates. This issue does NOT touch the Node/SQLite side (C3), does NOT implement cross-pack dedup or recency (C4), and does NOT change the desktop UI.

## Evidence motivating this slot
- CONFIRMED: [document_processor.py:349](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/document_processor.py#L349) — `doc_id = hashlib.sha256(canonical_path.encode()).hexdigest()[:16]`. Identity is derived from the **path**, not content. Two packs shipping the same content under different install paths index it twice; a file edited in place at the same path keeps the same `doc_id`.
- CONFIRMED: [vector_store.py:603-605](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/vector_store.py#L603) builds chunk ids as `f"{getattr(chunk, 'doc_id', None) or chunk.source}_{chunk.chunk_index}"` — a doc_id/path plus an index, not a content hash.
- CONFIRMED: [vector_store.py:650-663](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/vector_store.py#L650-L663) — `existing_ids = set(existing["ids"])` then `new_indices = [j for j, id_ in enumerate(ids) if id_ not in existing_ids]`: chunks whose id collides with an existing one are **silently skipped**, not replaced. Quote: `if new_indices: self.collection.add(...)` — only genuinely new ids are added; nothing is deleted first.
[...]

Issue #84, opened 2026-09-05
[Workstream E] PR 1 of 6: Package models and default packs into the Electron installer with an integrity manifest and a <=7 GB size budget


zaxbysauce:
**Roadmap:** E1 · **Workstream E: Hardening, CI and release**
**Sequence:** PR 1 of 6 (tracking slot; dependencies define execution order)
**Roadmap baseline:** 2026-09-05 · **Priority:** P0 · **Repo revision:** b2ddda0

## Outcome
The Electron installer contains every model/pack the app needs to answer tier-0 questions fully offline on first run, each staged asset is covered by a sha256 integrity manifest checked at startup, and the total footprint is measured and held at <= 7 GB. This does NOT certify runtime performance (E3) or pick which LLMs/embedder ship (A6/A5 ADRs) — it owns packaging mechanics, layout, and size accounting.

## Evidence motivating this slot
- Today's packaging is inconsistent across the three PyInstaller specs, none with a shared manifest or integrity check: [DocumentQAApp.spec:5](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/DocumentQAApp.spec#L5) bundles `('models','bundled_models')`; [AFOMIS.spec:18-22](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/AFOMIS.spec#L18-L22) bundles `bundled_models`, `models`, AND `seed_data` — CONFIRMED `seed_data/` does not exist at repo root; `app_gui.spec` ships no models datas entry at all.
- `installer.iss:43` does a blind recursive copy (`Source: "dist\AFOMIS\*"`) with no per-file manifest/hash — a corrupted model file ships silently.
[...]

PR #101, opened 2026-09-09
feat(desktop): B6 profile-scoped ingestion, backup and recovery on the SQLite store (issue #64)

### LINKED ISSUE #64: [Workstream B] PR 6 of 9: Profile-scoped ingestion, migrations, backup and recovery on the SQLite store

zaxbysauce:
Resolves #64.

## Root Cause

B6 scope was absent: the desktop SQLite store (B5) had no ingestion pipeline, no
profile-scoped layout, no backup/restore, no corruption recovery, and no cache
clearing. The latent root-cause defect this PR must not reproduce is the
stale-chunk class in the legacy Python path (`document_processor.py:347-349`):
document identity derived from a mutable file path, combined with an add path
that never deletes the prior revision, leaves superseded chunks/embeddings/FTS
rows live after a file is edited and re-ingested.

## Fix

Implements issue #64 end to end on the desktop store, content-hash-first:

- `desktop/main/backend/ingest/pipeline.ts` — IngestPipeline: extract
  (pdf/docx/xlsx/pptx/txt/md) → chunk (256 words / 100 overlap, CJK fallback) →
  embed (ONNX bge-small-en-v1.5 via transformers.js, 384-dim, dims-validated;
  without staged weights the ingest surface reports a staging diagnostic per
  request and documents stay stubbed; HashEmbedder is a dev/CI fixture via
  TRAININGAPP_DESKTOP_EMBEDDER=hash) → write via B5's frozen schema.
  Doc id = sha256(bytes); chunk id = sha256(docSha:idx:normalizedText);
  SHA-dedupe short-circuit before the write; delete-before-reingest of
  same-path stale revisions cascades chunks/embeddings/fts inside one
  BEGIN IMMEDIATE transaction; serialized single-writer queue.
- `desktop/main/backend/store/profiles.ts` — OS-user-scoped single profile
  DEFAULT at `<userData>/profiles/default/store.sqlite` (ADR
  `docs/adr/0006-profile-model.md`), opt-in named mode via
  `TRAININGAPP_PROFILE_NAME` (allowlisted), atomic legacy-layout rename
  migration; `profile.mode: "single"|"named"` default "single".
- `desktop/main/backend/store/backup.ts` — `createBackup`:
  `wal_checkpoint(TRUNCATE)` then copy to `<userData>/backups/<timestamp>/`;
  `restoreBackup` validates schema version/dims BEFORE any byte change.
- `desktop/main/backend/store/recovery.ts` — `PRAGMA integrity_check` on
  startup; corruption → Restore/Start-fresh prompt or auto-recovery, never a
  crash; non-corruption failures keep B5's log-and-continue degradation.
- `desktop/main/backend/store/document-surface.ts` — Clear Cache: close DB,
  delete `.sqlite`/`-wal`/`-shm`, re-init at the same path, keeps serving.
- Wiring: `/ingest/file` + `/ingest/batch` multipart routes (400 on missing
  part / >20 files), `ingest:progress {docId, phase, percent}` events to all
  windows, `desktop:store-backup` IPC, config keys
  `ingest.maxConcurrentFiles=2`, `ingest.chunkWordCount=256`,
  `ingest.chunkOverlapWords=100` (env-overridable, fail-safe).

- Defect class: store identity derived from mutable presentation data (a file
  path) instead of content, with an add path that never deletes the prior
  revision.
- Sweep result: 5 predicate hits, 5/5 dispositioned (trace 08a): the one
  path-hash site is the audited legacy Python surface (Workstream C, #69);
  the desktop port drops `generateDocId` entirely and the web_ui original is
  annotated `@deprecated — MUST NOT be used for store identity`.
- Guardrail: frozen regression test with independent identity recomputation
  (`desktop/src/__tests__/b6-stale-chunk.test.ts`, check C2) — demonstrated
  RED at base and biting via a reviewer mutation probe (disabling the
  delete-before-reingest loop fails C2).

## Tests

- Regression test: `bash repro-check.sh run` C1–C7 frozen checks → base RED /
  head GREEN on all seven (logs in trace `repro/`)
- Impacted suite: `npm --prefix desktop test` → 36 files / 258 tests PASS
- Lint/type/build/security checks: `npm --prefix desktop run compile` → 0
  TS errors; conformance host `node desktop/scripts/run-conformance-host.mjs`
  → PASS 10/10; Node↔Python store interop both directions → exit 0
- Bench: real-weights 200-page PDF 7.7s < 60s
  (`bench/ingest_bench_driver.mjs`; row in `bench/RESULTS.md`)
- Deferred-work scan: `scan-deferred.sh` -> clean

## Regression Protection

- 7 frozen b6 specs: conformance + negatives (C1), stale-chunk + independent
  id recomputation (C2), backup roundtrip + future-version refusal (C3),
  profile/config/progress/restart/chunker pins (C4), bench (C5), corruption
  recovery suite (C6), clear-cache (C7), plus b6-supplementary (dims
  mismatch, wrong vector count, per-file isolation, CJK end-to-end,
  docx/xlsx/pptx fixtures — found and fixed the xlsx `utils.sheet_to_csv`
  bug during this work).
- Independent reviewer mutation probes: (1) delete-before-reingest disabled →
  C2 fails; (2) restore copy moved before validation → C3 fails. Both
  reverted clean.
- Test drift: no existing test deleted, weakened, or skipped; frozen
  b2/b3/b4/b5 families unchanged and green; frozen API contract untouched
  (no-store stub semantics byte-identical).

|---|---|
| ADR for profile model (`docs/adr/0006-profile-model.md`) | ADR checked in; C4 pins single DEFAULT path, named-mode allowlist, default mode "single" |
| Ingest pipeline extract→chunk→embed→write, content-hash ids, delete-before-reingest | C2/C1 GREEN (`b6-stale-chunk.test.ts`, `b6-ingest-pipeline.test.ts`); pipeline.ts writeDocument |
| Progress events `ingest:progress {docId, phase, percent}` | C4 GREEN (`b6-profile-progress.test.ts`); host `onIngestProgress` → `webContents.send` |
| store/backup.ts checkpoint + copy + restore validates schema version | C3 GREEN (`b6-backup-recovery.test.ts`): roundtrip + future-version refusal |
| store/recovery.ts integrity_check, corruption → prompt not crash | C6 GREEN (`b6-corruption-recovery.test.ts`): detect, prompt, restore/fresh/auto |
| Clear Cache (close, delete db/-wal/-shm, re-init) | C7 GREEN (`b6-clear-cache.test.ts`): sidecars gone, keeps serving, empty corpus |
| 200-page PDF < 60s | C5 GREEN: 7.7s real-ONNX run; `bench/RESULTS.md` row `devstation / bge-small-en-v1.5 / 200 / 200 / 7.7 / 26 / pass` |
| Config keys 2 / 256 / 100 | C4 GREEN pins DEFAULT_INGEST_CONFIG and env overrides |

## Invariant Audit

- Frozen API contract (`contracts/api.openapi.yaml`): not touched —
  conformance host PASS 10/10; `DocumentInfo.id` remains the source path;
  >20-file batch → 400 preserved.
- Frozen store schema (`contracts/store.schema.sql`, B5): not touched — SQL
  matches the contract byte-for-byte; Node↔Python interop both directions
  exit 0; migrate ladder still a verified v1 no-op.
- BackendHost interface / b3 duck-type pin (host prototypes exactly
  [start, stop]): not touched — new capabilities are own properties/module
  functions; `b3-backend-selector` suite green.
- web_ui: not touched behaviorally (comment-only `@deprecated` annotation on
  `generateDocId`).

## Risk and Rollback

- Risk level: medium (new native deps: @huggingface/transformers ^3.8.1,
  pdfjs-dist ^4.10.38, mammoth ^1.12.2, xlsx ^0.18.5, jszip ^3.10.2;
  asarUnpack entries added)
- Rollback: single PR revert; legacy store layout restored by renaming
  `profiles/default/store.sqlite` back (migration is an atomic same-volume
  rename; no data rewrite)
- Residual risk: CI cannot run the AC2 timing check (LFS weights absent);
  recorded on weight-staged machines. Model weights staged out-of-band, per
  B4 convention.

## Waivers (or none)

none

## Merge status

AWAITING_USER_APPROVAL — awaiting explicit user approval; not merged.

PR head: 8e8aefe5808e991626fd8c9a2ce4f89c2febb046


zaxbysauce:
## Swarm PR review — findings and dispositions (run 2026-09-09-pr101)

Structured review executed on this PR (6 base lanes + 11 risk-family micro-lanes → 23 deduped candidates → independent reviewer validation → critic challenge). Final disposition:

### Fixed in 986a25c
| ID | Sev | Finding | Fix |
|---|---|---|---|
| PRR-002 | MEDIUM | `createBackup` copied directly to the final timestamped path — a crash mid-copy could leave a partial `store.sqlite` that `latestBackup()` would later select for restore | Stage to `<ts>/store.sqlite.part`, then atomic rename; `.part` siblings are invisible to restore selection; pinned in `b6-review-hardening.test.ts` |
| PRR-005 | MEDIUM | Unbounded extraction: zip-based parsers (docx/xlsx/pptx) had no decompressed-size bound, directory ingest had no per-file size cap, extracted text was unbounded | Env-overridable caps enforced per-file in every mode: `TRAININGAPP_INGEST_MAX_FILE_BYTES` (60MB), `TRAININGAPP_INGEST_MAX_ZIP_BYTES` (512MB, central-directory pre-scan), `TRAININGAPP_INGEST_MAX_TEXT_CHARS` (8M); pinned in tests |
| PRR-004 | LOW | `modelIdRecorded` was set before COMMIT, so a rolled-back first write left `meta.embedding_model_id` empty for the pipeline's lifetime | Flag now set only after COMMIT succeeds; pinned in tests |
| PRR-016 | MEDIUM | clear-cache sidecar removal failures were silently swallowed; the comment claimed re-init "fails loud" (it does not — openStore doesn't check sidecars) | Failures now logged loudly; comment corrected |
| PRR-010 | MEDIUM | tracked vitest-cache artifact (`node_modules/.vite/vitest/…/results.json`) churned by 31d7369 | Untracked; `node_modules/` ignore now covers it |
| PRR-009/017 | MEDIUM/LOW | ADR didn't document that a failed migration aborts launch, nor the downgrade data-invisibility consequence | ADR Consequences updated |
| PRR-013/014 | LOW/MEDIUM | missing-file-part 400 asserted status only; restore onto a NON-EMPTY store untested | New additive `b6-review-hardening.test.ts` pins the 400 body detail and restore-replaces-live-store |

### Dispositioned without code change (reviewer/critic adjudicated)
- **PRR-001** (clear-cache vs in-flight ingest closes the store under the pipeline): contained by per-file failure isolation — visible error, no corruption, no data loss; queue-drain is B8 scope. Follow-up noted.
- **PRR-003** (restore copy not atomic): restore only ever runs on an already-corrupt store; the backup source survives any mid-copy crash and the next launch retries — no incremental loss (critic ruling).
- **PRR-006** (throwing ROLLBACK leaves connection unusable): requires catastrophic I/O to trigger; errors are visible per-file; restart cures (critic: LOW).
- **PRR-007** (dev-server backups dir divergence): test/dev-only entrypoint; tests pass explicit `storeBackupsDir`; no user impact.
- **PRR-008** (locked-but-healthy store degrades without retry): deliberate b5-pinned degrade design; single-instance lock removes the common case; every ingest returns an honest error (critic: LOW).
- **PRR-011**: `onnxruntime-web` dev-prerelease pin is upstream's own; never loaded on Node (runtime uses onnxruntime-node). Inert.
- **PRR-012** (no npm audit in security.yml): pre-existing CI gap, workflow untouched by this PR. Follow-up.
- **PRR-015** (directory ingest follows symlinks): no privilege boundary crossed (user's own local store). Follow-up hardening.
- **PRR-018** (parentless corruption dialog): standard for a pre-window fatal prompt; functional.
- **PRR-019** (`/health` engine_ready hardcoded): pre-existing at origin/master (git-verified), out of PR scope. Follow-up.
- **PRR-020** (3 exported-but-unused symbols): accepted non-blocking nits.
- **PRR-021**: PR body inaccuracies (test count, embedder fallback wording, dep versions) corrected in the body.
- **PRR-022/023**: identity/content covered by `b6-stale-chunk` content assertions; IPC forwarder wiring accepted (surface callback is tested).

### Rejected with evidence
asarUnpack-missing-sqlite-natives (CRITICAL claim rejected: origin/master had no asarUnpack and electron-builder smartUnpack unpacks native addons by default); backup-sequence modulo-wrap infinite loop (physically unreachable); decodeXmlEntities RangeError (per-file failure isolation by design); windows-only desktop CI (deliberate — Windows is the shipping platform).

Desktop suite after the fixes: **258/258 across 36 files**. Frozen b6 specs untouched (all new pins are additive).

zaxbysauce:
## Swarm review: PR #101 (feat(desktop): B6 profile-scoped ingestion, backup and recovery)

6-lane adversarial council (correctness, security, dependency/deployment, docs/intent-vs-actual, tests/falsifiability, durability/architecture) + independent orchestrator verification of every high-impact claim against source at `c7f3439`. Findings below are ones I personally re-read in the tree, not raw agent output.

### PR intent
Implement issue #64 end-to-end on the desktop SQLite store: content-hash-first ingestion (extract→chunk→embed→write), single OS-user profile with opt-in named mode, backup/restore, corruption recovery, and cache clearing — while specifically avoiding the legacy Python path's stale-chunk defect (path-derived doc identity + no delete-on-reingest).

### Intended vs. actual — verified
- **Content-hash identity + delete-before-reingest in one transaction**: CONFIRMED. `pipeline.ts:203` (`docId = sha256(bytes)`), `pipeline.ts:313-362` wraps cascade-delete + insert in one `BEGIN IMMEDIATE`/`COMMIT`, with `ROLLBACK` on error. This is the PR's central claim and it holds.
- **`b6-stale-chunk.test.ts` mutation-kill claim**: CONFIRMED — independently traced: removing the delete loop at `pipeline.ts:317-328` would leave 2 doc rows where the test asserts 1, and leave stale-marker text where the test asserts zero. This is a genuine regression guard, not test theater.
- **Progress events, multipart routes/400s, config defaults, `generateDocId` dropped from desktop, web_ui `@deprecated` comment-only, ADR-0006 vs. `profiles.ts`, second-commit scope**: all CONFIRMED as described (see docs-lane detail). Minor doc nits only: "config keys" phrasing implies a dotted settings key that doesn't exist (only env vars do); backup dir format description in README omits the sequence suffix.

### Confirmed findings (ship-blocker tier)

**1. Vulnerable `xlsx` (SheetJS) 0.18.5 in the untrusted ingest path — CVE-2023-30533, and it cannot be fixed by a normal version bump.**
`desktop/package.json:35` pins `"xlsx": "^0.18.5"`. This is the last version SheetJS published to the npm registry; the fix (0.19.3+) was only ever distributed via `cdn.sheetjs.com`, never republished to npm — so no `^`/`~` range against the registry can resolve to a patched build. Attacker-controlled ingested `.xlsx` bytes reach this parser directly and unsandboxed at `extractors.ts:119-126` (`XLSX.read(fs.readFileSync(filePath), ...)`), registered for real ingestion (not just tests) at `extractors.ts:27,180`. Recommend: point the dependency at the SheetJS CDN tarball, vendor a patched build, or swap parsers before merge.

**2. `text-chunker.ts:177` — the 256-word chunk-size invariant is broken for any oversized sentence that isn't the very first one.**
```ts
if (sentenceUnitCount > this.chunkSize && currentChunkSentences.length === 0) { ... split logic ... }
```
Independently verified: once any sentence has already been buffered (true almost immediately into any real document — overlap-seeding rarely leaves the buffer empty), a later oversized sentence skips this branch entirely, falls through to the plain append path at `text-chunker.ts:213-229`, and is pushed into a chunk with **no length check at all** — neither there nor in the final flush at `text-chunker.ts:232-234`. A single unpunctuated/minified/CJK-fallback-eligible sentence anywhere but the document's first position produces one chunk far exceeding the configured word limit, silently. Falsifiable with a 2-sentence fixture: short sentence 1, 5000-word sentence 2 → chunk 2 is untouched at 5000 words. Recommend a regression test with the long sentence NOT first, and remove the `currentChunkSentences.length === 0` guard (or apply the same split path regardless of buffer state).

**3. `store/backup.ts:41` — `wal_checkpoint(TRUNCATE)`'s result is discarded; `restoreBackup` overwrites the live store with a non-atomic copy and no pre-replace safety snapshot.**
`checkpoint()` calls `.get()` on the pragma and never inspects `busy`/`checkpointed` — a busy/partial checkpoint (e.g. blocked by a concurrent reader) is invisible, and `createBackup` proceeds to `fs.copyFileSync` regardless (`backup.ts:56-66`). Independently verified `restoreBackup` at `backup.ts:138`: schema/dims validation genuinely precedes any write (matches the doc comment), but the replace itself is a direct `fs.copyFileSync(opts.backupPath, opts.dbPath)` onto the live file — no temp-write-then-rename, no safety copy of the live store taken first. A crash mid-copy during restore leaves the active store truncated with no automatic fallback.
Context that meaningfully narrows real-world impact: independently grepped the entire `desktop/` tree for `journal_mode` — **zero matches**. WAL is never actually enabled anywhere in this codebase, so the header comment's premise ("after checkpoint the -wal file is empty, so copying the main file alone is a complete snapshot") doesn't describe the DB's actual configured mode. This doesn't make the restore-atomicity gap safe, but it does mean the checkpoint-result-ignored issue is likely closer to dead code than an active corruption vector today — worth a reviewer sanity check on whether WAL was intended and is a missing `PRAGMA journal_mode=WAL`, or the comments are aspirational.

**4. `store/profiles.ts` (migration) and `backup.ts`/`recovery.ts` have no mutual exclusion with the ingest pipeline's single-writer queue.**
`pipeline.ts`'s `queue`/`enqueue()` (`pipeline.ts:87,103-107`) only serializes calls made through the pipeline itself. `index.ts`'s `createStoreBackup` calls `createBackup(store, backupsDir)` directly on the shared store handle, entirely outside that queue (grepped `backup.ts`/`recovery.ts` for any `enqueue`/`queue` reference — none). A backup, restore, or Clear Cache can run concurrently with an in-flight ingest with no coordination. Concretely: `document-surface.ts:97` nulls the pipeline reference without draining in-flight work; a resumed `ingestOne` past an `await` (extract/embed) later touches the now-closed `db` handle and throws. Not silent data corruption in the traced path, but unsynchronized and untested (no race coverage found in `b6-clear-cache.test.ts`).

### Confirmed findings (should-fix, non-blocking)

**5. `ingest/embedder.ts:195` — the "fail loud, never mid-write crash" contract has a bypass.**
```ts
const staged = OnnxEmbedder.stagedModelDir([explicit, ...]) ?? (explicit !== undefined && explicit.length > 0 ? explicit : null);
```
Independently verified: when `TRAININGAPP_EMBEDDING_MODEL_DIR` is **explicitly set but invalid** (e.g. an unresolved Git-LFS pointer, caught by `isValidModelDir`'s ≥10MB check), `stagedModelDir` correctly returns `null` for every candidate — but the `??` fallback then hands back the *invalid* `explicit` path anyway instead of `null`, so the `ModelNotConfiguredError` guard at lines 196-202 never fires. `OnnxEmbedder` gets constructed over the bad path, and failure only surfaces later inside transformers.js/onnxruntime mid-ingest — exactly the failure mode this module's own doc comment (`embedder.ts:178-181`) says it prevents.

**6. `electron-builder.yml` `asarUnpack` covers the embed runtime but not the store's native deps** — `better-sqlite3`, `sqlite-vec`, `node-llama-cpp` are absent from the new `asarUnpack` list, so a packaged (asar) build still cannot open the SQLite store at all (native `.node`/DLL loading from inside asar is a hard Electron limitation). This is explicitly disclosed pre-existing scope (`sqlite-store.ts:12-15`, `README.md:147-149` both point at #84 as the owner), not something this PR silently broke — but worth a merge-time question: should #84 (packaged native-addon unpacking) block shipping B6, given B6 is nominally "on the SQLite store" and a packaged install can't use it yet?

**7. `profiles.ts:85` legacy migration renames only `store.db`, never `-wal`/`-shm` sidecars.** Practically self-limiting given finding #3's WAL-never-enabled observation (sidecars likely never exist in current runtime behavior), but if WAL is ever turned on later this silently drops data on migration, and the failure would be invisible (the cleanup `rmdirSync` at `profiles.ts:88` would fail non-empty and is swallowed by a bare `catch`).

**8. `b6-corruption-recovery.test.ts` never exercises the actual `PRAGMA integrity_check` malformed-but-openable-file branch** — its `corruptFile()` fixture (`Buffer.alloc(4096, 0xde)`) isn't a valid SQLite file at all, so `better-sqlite3`'s *open* throws before the pragma ever runs; all 5 tests in the file only cover the open-failure path (`recovery.ts:41-56` catches both under one `{ok:false}` shape). Deleting the `PRAGMA integrity_check` call would not fail any of these tests. Needs a fixture with a valid header + corrupted interior pages to actually pin the pragma.

### Minor / informational
- `/ingest/batch`'s >20-file check (`server.ts:526`) runs after all parts are buffered into memory, not before — bounded overall by the pre-existing 60MB body cap, so not unbounded DoS, just enforcement-ordering.
- Long-sentence split path uses a halved overlap (`text-chunker.ts:198-201`) vs. the full configured overlap used on the normal sentence-boundary path — undocumented spec deviation, low impact.
- `pipeline.ts`: the initial dedupe `SELECT` (`pipeline.ts:304-306`) sits outside the transaction's `try`, so a DB-level exception there could make `ingestFile` reject a Promise despite its documented "never throws" contract.
- `bench/RESULTS.md`'s reported `chunks == pages` (200/200) is numerically inconsistent with a whole-document overlapping chunker on ~36k words of synthetic text — recommend an independent re-run before trusting the 7.7s/26-emb/s figures.
- `bench/append_results.py` has no file-locking/atomic-write around `RESULTS.md` (dev tooling only, low severity).
- `desktop:store-backup` IPC handler leaks raw `err.message` (absolute paths) to any future renderer caller (`index.ts:84-86`), unlike the HTTP routes which return a static error string. Currently unreachable — no preload bridge exists yet — but should be fixed before B9 wires one, not after.
- Pre-existing, out of this PR's diff: `electron-builder.yml`'s `files:` list (`dist/**/*`, `package.json`) predates this PR — whether `node_modules` is packed at all is a question for #84's scope, not this PR's.

### Test/coverage gaps
- No test for concurrent ingest-vs-backup/clear-cache (finding #4).
- No test that actually corrupts SQLite interior pages rather than replacing the whole file (finding #8).
- No test for the long-sentence-not-first chunk overflow (finding #2).
- No test for the embedder invalid-explicit-path bypass (finding #5).

### Verdict
The PR's headline claim — genuinely fixing the stale-chunk defect class with a real, mutation-tested regression guard — holds up under independent re-verification. The remaining findings are real but mostly narrow: one supply-chain issue that needs a real fix before merge (#1, unpatchable via registry bump), one correctness bug with a concrete repro (#2), and a cluster of durability/coordination gaps (#3/#4) whose blast radius is reduced (not eliminated) by WAL never being enabled in this codebase today.

**Merge recommendation: request changes**, gated on #1 (xlsx CVE — supply-chain, not optional) and #2 (chunk-size invariant — concrete, cheaply fixed, cheaply tested). #3–#8 are legitimate follow-up-worthy but not individually blocking given the WAL-disabled mitigating context; recommend a fast-follow issue tracking #3/#4/#5/#8 rather than re-scoping this PR further.

---

https://claude.ai/code/session_01GwgeD1H4UfKaz5ok5GkP7x

Issue #71, opened 2026-09-05
[Workstream C] PR 4 of 9: Version precedence, cross-pack chunk dedup, and recency-weighted ranking


zaxbysauce:
**Roadmap:** C4 · **Workstream C: Knowledge packs and freshness**
**Sequence:** PR 4 of 9 (tracking slot; dependencies define execution order)
**Roadmap baseline:** 2026-09-05 · **Priority:** P1 · **Repo revision:** b2ddda06e86b63f81bdbedce220a3a2357bc9852

## Outcome
When this closes, retrieval prefers newer content when two active packs (or a user doc and a bundled doc) contain overlapping or superseding material, and citations expose which pack/version an answer came from. This issue does NOT install/remove packs (C2/C3 own that) and does NOT touch the chat UI badge (C5 owns the grounding surface, though this issue's ranking output feeds it).

## Evidence motivating this slot
- CONFIRMED: [utils.py:4-24](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/utils.py#L4-L24) `rrf_fuse(results_list, k=60)` implements Reciprocal Rank Fusion with `score = 1/(k+rank)`; this is the fusion stage this issue's recency prior must apply AFTER (recency multiplies the fused RRF score, it does not replace it or feed into `k`), matching the existing web_ui plan (`rag-orchestrator.ts` RRF is analogous per audit_report.md 2.1).
- CONFIRMED: [vector_store.py:954-1047](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/vector_store.py#L954-L1047) `delete_document` proves superseded-pack chunk removal is mechanically available once C2 calls it on supersede — this issue's "superseded excluded from ranking" requirement is largely satisfied for free if C2 truly deletes on supersede; this issue's real work is the case where TWO packs are simultaneously ACTIVE (not one superseding the other) and their chunks overlap (e.g. a bundled doc and a later `user` upload of the same content).
[...]

Issue #63, opened 2026-09-05
[Workstream B] PR 5 of 9: Prove Node/Python sqlite-vec interoperability and freeze the store schema


zaxbysauce:
**Roadmap:** B5 · **Workstream B: Electron shell and desktop backend**
**Sequence:** PR 5 of 9 (tracking slot; dependencies define execution order)
**Roadmap baseline:** 2026-09-05 · **Priority:** P0 · **Repo revision:** b2ddda0

## Outcome
Closing this freezes `contracts/store.schema.sql`, proves a fixture SQLite DB written by `better-sqlite3` (Node) is read byte-identically by Python's sqlite-vec binding and vice versa, and pins the sqlite-vec extension version. Does NOT certify ingestion, migrations, or backup/recovery (B6/#64) — only that the file format is genuinely shared, the load-bearing assumption for "one store, both backends."

## Evidence motivating this slot
- `implementation_plan.md:27,42` states the target ("SQLite + sqlite-vec + FTS5 in ONE file per profile... the Python side can read the same file") as a PLAN assertion, not yet demonstrated. `critic_sol_verdict.md` risk 4: "The 'same SQLite file' promise is unproven across better-sqlite3, Python and browser; sqlite-vec is explicitly pre-v1. Mitigation: freeze extension/schema versions and run byte-identical interop fixtures before packtool."
- Today's stores are NOT shared: Python uses ChromaDB (`vector_store.py`, per `audit_report.md` section 1) and the browser uses EdgeVec HNSW + FlexSearch + IndexedDB (`audit_report.md` section 2.1) — CONFIRMED via the audit's citations. This is greenfield schema design.
- [web_ui/src/lib/storage/profile.ts:20-29](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/web_ui/src/lib/storage/profile.ts#L20-L29) — CONFIRMED: `StorageDbNames { documents, vectorMapping, vector, keyword }` — four separate IndexedDB stores today; the new schema folds these into one SQLite file's tables.
[...]

PR #102, opened 2026-09-10
feat(desktop): B7 hybrid retrieval with RRF fusion, worker-thread reranking, and eval parity (issue #65)

### LINKED ISSUE #65: [Workstream B] PR 7 of 9: Hybrid retrieval (sqlite-vec + FTS5 -> RRF -> reranker) with eval parity

zaxbysauce:
Closes #65

[Workstream B] PR 7 of 9 — Hybrid retrieval (sqlite-vec + FTS5 → RRF → reranker) with eval parity. Replaces the B3 stub on the desktop backend with the full hybrid pipeline, validated end-to-end under the issue-tracer contract (`.agents/issue-traces/65-hybrid-retrieval-rrf-reranker/`).

## Root Cause

Issue #65 is a FEATURE roadmap slot: the desktop backend had no retrieval implementation. `StubEngine.search()` (desktop/main/backend/engine.ts) returned a single hardcoded `desktop-stub` row on every path — `/search` (server.ts → engine.search) and the retrieval step inside `/ask`//ask/stream (llama-engine.ts returned `sources: []`). None of the Required-scope capabilities (vector KNN, FTS5, fusion, reranking, floor) existed anywhere in desktop/main. The implementation also surfaced a hidden runtime constraint: onnxruntime-node 1.21 **aborts the whole process (exit 134)** when the same ONNX module is used from the main thread and a user worker thread (empirically proven: main embed → worker rerank → main embed = deterministic `OnFatalError`; main-only embeds stable).

## Fix

Minimal, complete implementation of the Required scope:

- `desktop/main/backend/retrieval/hybrid.ts` — vec0 KNN leg ∪ FTS5 BM25 leg (legK = topK × candidateMultiplier), reciprocal-rank fusion `1/(rrfK + rank + 1)` deduped by chunk id, `recencyWeight()` hook (returns 1.0; the C4/#71 extension point), ettin cross-encoder rerank over the fused window with scores replacing RRF scores, calibrated relevance floor, topK slice; FTS5 query sanitization (operators/quotes/stars dropped, tokens quoted — never throws); empty store returns `[]` without touching the reranker; per-query reranker failure degrades to fused ordering (logged once).
- `desktop/main/backend/retrieval/config.ts` — frozen defaults `topK=10, candidateMultiplier=3, rerank=true, rrfK=60, relevanceFloor=0.569387` (re-calibrated — NOT the browser's un-revalidated 0.2; see `docs/adr/0007-relevance-floor-calibration.md`) with per-key `TRAININGAPP_RETRIEVAL_*` env overrides and per-key fallback on invalid values.
- `desktop/main/backend/retrieval/reranker.ts` + `rerank-worker.ts` — **single-thread ORT ownership**: ALL onnxruntime work (embedding AND reranking) lives in ONE worker thread; the main thread uses `WorkerEmbedder`/`WorkerReranker` proxies exclusively. This is a deliberate strict superset of the plan's "reranker in worker" (documented at index.ts and in the worker header) — the planned "embedder on main" wiring deterministically aborts the process.
- Wiring — `EngineSurface.attachRetrievalSurface` seam (mirrors B6's document surface); `NodeBackendHost.start()` resolves the embedder once, wraps it in `WorkerEmbedder` when a reranker exists, attaches the retrieval surface after the store surface; `StubEngine`/`LlamaEngine` ground prompts from retrieved chunks and populate sources/context_length; detaching the surface restores the frozen `desktop-stub` conformance row (asserted by C5).
- Bench — `desktop-retrieval` surface registered in `bench/append_results.py` (5 points); committed `bench/retrieval_bench_driver.mjs`; `bench/RESULTS.md` § Desktop retrieval (B7): `devstation | bge-small-en-v1.5 | ettin-reranker-32m-v1 | 10 | 3 | p50=364 | p95=426 | max=691 | pass` (budget p95 ≤ 1500 ms).

- Defect class: native ONNX inference loaded/run outside the single owning worker thread (aborts the process under onnxruntime-node's single-thread ORT ownership).
- Sweep result: 11 hits across 4 repo-wide predicates (ONNX API calls 4, native-stack imports 3, worker spawns 1, stub sentinel 3) — every hit dispositioned to the sanctioned owner files or the intentional frozen stub sentinel; see `.agents/issue-traces/65-hybrid-retrieval-rrf-reranker/08a-recurrence-sweep.md`.
- Guardrail: permanent vitest source scan (`desktop/src/__tests__/b7-ort-ownership-guardrail.test.ts`, runs in the standard suite) that fails when ONNX APIs/imports appear outside the two owner files or a second worker appears. Demonstrated RED (injected main-thread ONNX import → `engine.ts:276` flagged) then GREEN (reverted): transcripts in the trace `repro/guardrail-{red,green}.log`.

## Tests

- Regression test: `cd desktop && npx vitest run` → 43 files / **291 passed** (0 failures, 0 unhandled errors)
- Frozen acceptance specs: `run-check.sh C1` → 12 passed; `C2` → 8 passed; `C4` → 2 passed; `C5` → 3 passed; `C6` → 3 passed (all RED/ERROR at the frozen pre-fix checkpoint)
- Drivers: `C3` → `C3 PASS floor=0.569387` (re-derived through the production reranker worker, Δ < 0.001); `C7` → `C7 PASS recall@5=0.640 mrr=0.384` (errors=0, honest parity — a vacuous first pass with 55/56 errors was caught during validation and re-run after the ORT fix); `C8` → `C8 PASS p95=517ms` (20 store-backed round trips); `C9` → `C9 PASS` (10/10 conformance, GREEN-to-GREEN preserving)
- Lint/type/build: `npm --prefix desktop run compile` (tsc) clean; `black --check` + `flake8 --max-line-length=100` clean on `bench/append_results.py`
- Deferred-work scan: `scan-deferred.sh` → clean (base=origin/master)

## Regression Protection

- 7 new spec files under `desktop/src/__tests__/b7-*.test.ts` (fusion math, config freezing + env overrides, worker isolation via event-loop tick counting, host wiring incl. detach, recency-hook inertness, FTS sanitization fuzzing, ORT-ownership guardrail)
- Adversarial: worker-dispose rejects in-flight jobs (idempotent); operator-heavy FTS queries never throw; per-query reranker crash degrades to RRF-only ordering
- Test drift review: two sanctioned checkpoint AMENDs, neither weakening a check — seq 12 `CHECK_WRONG` (the original C1 reference embed made its own assertion unsatisfiable by construction; amended spec keeps the exactly-once semantic) and seq 13 `FORMAT_ONLY` (the `checkpoint-tree-id:` binding line the phase 2.5 validator demanded; acceptance table byte-identical). `verify-checkpoint` exits 0 on all 13 rows.

| Acceptance criterion | Evidence |
|---|---|
| AC1 hybrid.ts: vec0 ∪ FTS5 → RRF k=60 → recency hook → worker rerank → floor → topK | C1 `Tests 12 passed (12)`; hybrid.ts pipeline |
| AC2 config keys + calibrated floor | C2 `Tests 8 passed (8)`; config.ts frozen defaults; ADR-0007 |
| AC3 floor re-calibration documented (≠ 0.2) | C3 `C3 PASS floor=0.569387` (production worker re-derivation within ±0.05) |
| AC4 reranker in dedicated worker thread, never main | C4 `Tests 2 passed (2)` (event-loop tick counter ≥40 during in-flight job); single-thread ORT ownership wiring + guardrail |
| AC5 /search + /ask//ask/stream wired, B3 stub replaced (detach restores stub) | C5 `Tests 3 passed (3)` over real loopback HTTP |
| AC6 recencyWeight() returns 1.0 (C4/#71 hook inert) | C6 `Tests 3 passed (3)` over 40+ input shapes |
| AC7 eval parity vs web_ui baseline | C7 `C7 PASS recall@5=0.640 mrr=0.384` errors=0 vs baseline 0.640/0.324 (recall@1 0.180≥0.100, recall@3 0.420≥0.340) |
| AC8 p95 ≤ 1.5 s + bench/RESULTS.md row | C8 `C8 PASS p95=517ms`; RESULTS.md row p95=426 ms (committed driver) |
| AC9 /search conformance | C9 `CONFORMANCE: PASS 10/10` |

## Invariant Audit

The desktop backend contract docs (`desktop/README.md`, ADRs 0005–0007) plus the cross-workstream frozen contracts:

- B3 `/search` response shape + stub-row detach behavior: touched, preserved — C5 asserts the stub row returns verbatim on `attachRetrievalSurface(null)`; C9 conformance 10/10.
- B6 store/ingest surfaces: not touched semantically — embedder gains a read-only `weightsDir` getter; `attachStoreSurface` unchanged; full desktop suite green.
- sqlite-vec store schema (ADR-0005, frozen in #63): not touched — read-only KNN queries via the frozen interop API.
- Eval harness contract (A4): not touched — `eval/runner.py` consumed as-is; baseline JSON untouched.

## Risk and Rollback

- Risk level: medium (new runtime path on the desktop backend; reranker weights are operator-staged — without them retrieval degrades to RRF-only with no floor, by design and logged once)
- Rollback: revert commit 55b1646 (single commit, no migrations; the store is untouched)
- Residual risk: the ORT cross-thread abort is enforced by guardrail test + worker-only wiring; a future main-thread ONNX call would be caught by the suite. Cross-thread ORT behavior is a property of onnxruntime-node 1.21 and should be re-verified on upgrade (guardrail + C7/C8 cover it).

## Scope-outs (no follow-up issue required)

- Sidecar-mode hosts intentionally keep B3 behavior (hybrid retrieval requires the in-process Node backend; the B6 precedent).
- `recencyWeight()` stays 1.0 — the Knowledge Pack recency hook is C4 / issue #71, which already tracks it.
- The bench row is measured on `devstation` (machine-tagged provenance; the reference-laptop row lands via the RESULTS.md runbook when an operator runs it there).

## Review pass (post-publication hardening)

A full swarm review of this PR validated 17 findings (1 initially Critical,
re-banded MEDIUM follow-up after two critic passes; 9 MEDIUM; 5 LOW; 4
rejected with evidence). The actionable fixes landed in efafa40: reranker
failure latch, non-finite score guard, 8000-char query cap, env-int upper
bounds, worker dispose/exit settlement, bench-driver fetch deadlines, and a
6-test additive hardening spec. Suite grew 291 -> 297 (green with weights and
in the CI-equivalent no-weights environment). The MEDIUM follow-up (an
in-repo eval-parity floor, analogous to bench-floors) is tracked in a
follow-up issue filed at merge.

## Waivers (or none)

none.

## Merge status

User approval to resolve review findings and merge was recorded 2026-09-10 (trace 10b-merge-approval.md, bound to head efafa407d4a8e3314ae286005d816179ec5eec93); merge pending CI green on that head.

---

PR record: https://github.com/ZaxbyHub/trainingapp/pull/102 · head SHA: efafa407d4a8e3314ae286005d816179ec5eec93 (head 3: PR-review hardening; head 2 e23b460: CI no-weights degrade fix; head 1 55b1646: initial implementation) · base: master · issue comment: https://github.com/ZaxbyHub/trainingapp/issues/65#issuecomment-5611770772
PR head: efafa407d4a8e3314ae286005d816179ec5eec93

### DIFF
--- a/bench/RESULTS.md
+++ b/bench/RESULTS.md
@@ -229,3 +229,20 @@ synthetic N-page PDF generated in-process. Budget: 200 pages in < 60 s
 | machine | model | pages | chunks | total_s | embeddings_per_second | outcome |
 |---|---|---|---|---|---|---|
 | devstation | bge-small-en-v1.5 | 200 | 200 | 7.7 | 26 | pass |
+
+## Desktop retrieval (B7)
+
+Hybrid desktop retrieval (issue #65): sqlite-vec KNN leg + FTS5 BM25 leg ->
+reciprocal-rank fusion (k=60) -> ettin-reranker-32m-v1 cross-encoder over the
+fused top-30 in the rerank worker thread, measured by
+`bench/retrieval_bench_driver.mjs`: boot the compiled desktop host with real
+weights, ingest the 8-doc eval corpus over /ingest, then 20 sequential timed
+POST /search round trips over the in-corpus eval questions (loopback HTTP, so
+each sample is the full stack: query embed -> vec0 + FTS5 legs -> RRF fuse ->
+rerank -> floor/slice; the cold first query — reranker worker spawn + model
+load — is part of the measured distribution). Budget: p95 <= 1500 ms
+(issue #65 acceptance).
+
+| machine | embedder | reranker | topk | multiplier | p50_ms | p95_ms | max_ms | outcome |
+|---|---|---|---|---|---|---|---|---|
+| devstation | bge-small-en-v1.5 | ettin-reranker-32m-v1 | 10 | 3 | 364 | 426 | 691 | pass |
--- a/bench/append_results.py
+++ b/bench/append_results.py
@@ -37,6 +37,7 @@ SECTION_HEADINGS = {
     "onnx-embed": "## ONNX embed/rerank cost",
     "onnx-rerank": "## ONNX embed/rerank cost",
     "desktop-ingest": "## Desktop ingest (B6)",
+    "desktop-retrieval": "## Desktop retrieval (B7)",
 }
 
 # surface -> header cells of the table that accepts the surface's rows. The
@@ -75,6 +76,17 @@ HEADER_CELLS = {
         "chunks",
         "total_s",
         "embeddings_per_second",
+        "outcome",
+    ],
+    "desktop-retrieval": [
+        "machine",
+        "embedder",
+        "reranker",
+        "topk",
+        "multiplier",
+        "p50_ms",
+        "p95_ms",
+        "max_ms",
         "outcome",
     ],
 }
@@ -123,6 +135,17 @@ REQUIRED_KEYS = {
         "total_s",
         "outcome",
     ),
+    "desktop-retrieval": (
+        "surface",
+        "machine",
+        "embedder",
+        "reranker",
+        "topk",
+        "multiplier",
+        "p50_ms",
+        "p95_ms",
+        "outcome",
+    ),
 }
 
 # Provenance keys every row (pass or fail) must carry.
@@ -139,6 +162,13 @@ IDENTITY_CELLS = {
     "onnx-embed": (0, 1),  # machine, model
     "onnx-rerank": (0, 1),  # machine, model
     "desktop-ingest": (0, 1),  # machine, model (mirrors onnx-embed)
+    "desktop-retrieval": (
+        0,
+        1,
+        2,
+        3,
+        4,
+    ),  # machine, embedder, reranker, topk, multiplier
 }
 
 REGISTRY_HEADING = "## Machine registry"
@@ -180,6 +210,18 @@ def row_cells(row: dict) -> list:
             row.get("model", ""),
             fmt(row.get("top15_ms")),
             fmt(row.get("top30_ms")),
+            row.get("outcome", ""),
+        ]
+    if row.get("surface") == "desktop-retrieval":
+        cells = [
+            row.get("machine", ""),
+            row.get("embedder", ""),
+            row.get("reranker", ""),
+            fmt(row.get("topk")),
+            fmt(row.get("multiplier")),
+            fmt(row.get("p50_ms")),
+            fmt(row.get("p95_ms")),
+            fmt(row.get("max_ms")),
             row.get("outcome", ""),
         ]
     elif row.get("surface") == "desktop-ingest":
--- a/bench/retrieval_bench_driver.mjs
+++ b/bench/retrieval_bench_driver.mjs
@@ -1,0 +1,237 @@
+#!/usr/bin/env node
+// B7 bench driver (issue #65): measure the SHIPPING desktop hybrid retrieval
+// pipeline end-to-end — boot the compiled desktop host (real bge-small embed +
+// ettin-reranker-32m-v1 rerank in the worker thread), ingest eval/corpus over
+// /ingest, then time sequential POST /search round trips over the in-corpus
+// eval questions and emit one JSON row compatible with
+// `bench/append_results.py` (surface=desktop-retrieval). Timings are full
+// round trips over loopback HTTP, so they include the whole stack: query
+// embed -> vec0 KNN + FTS5 legs -> RRF fuse -> rerank top-30 -> floor/slice.
+// The first (cold) query — reranker worker spawn + model load — is part of
+// the measured distribution, matching real interactive usage.
+//
+// Requires: `npm --prefix desktop run compile` and staged weights
+// (models/bge-small-en-v1.5 required; models/ettin-reranker-32m-v1 optional —
+// without it the row records reranker=none and the budget still applies).
+//
+// Usage (from repo root):
+//   node bench/retrieval_bench_driver.mjs --queries 20 [--json row.json]
+// Machine tag: BENCH_MACHINE_TAG env (default devstation).
+import fs from 'node:fs';
+import path from 'node:path';
+import { performance } from 'node:perf_hooks';
+import { pathToFileURL, fileURLToPath } from 'node:url';
+import { spawn, spawnSync } from 'node:child_process';
+
+const repoRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..');
+const IS_WIN = process.platform === 'win32';
+const P95_BUDGET_MS = 1500;
+
+function parseArgs(argv) {
+  const args = {};
+  for (let i = 0; i < argv.length; i += 1) {
+    switch (argv[i]) {
+      case '--queries': args.queries = Number.parseInt(argv[++i], 10); break;
+      case '--json': args.json = argv[++i]; break;
+      default: throw new Error(`unknown argument: ${argv[i]}`);
+    }
+  }
+  if (!Number.isInteger(args.queries) || args.queries < 1) {
+    throw new Error('--queries is required and must be a positive integer');
+  }
+  return args;
+}
+
+function percentile(values, pct) {
+  const ordered = [...values].sort((a, b) => a - b);
+  if (ordered.length === 1) return ordered[0];
+  const pos = ((ordered.length - 1) * pct) / 100;
+  const low = Math.floor(pos);
+  const high = Math.min(low + 1, ordered.length - 1);
+  const frac = pos - low;
+  return ordered[low] * (1 - frac) + ordered[high] * frac;
+}
+
+function loadInCorpusQuestions() {
+  const raw = fs.readFileSync(path.join(repoRoot, 'eval', 'questions.jsonl'), 'utf8');
+  const rows = [];
+  for (const line of raw.split(/\r?\n/)) {
+    const trimmed = line.trim();
+    if (!trimmed || trimmed.startsWith('#')) continue;
+    const row = JSON.parse(trimmed);
+    if (row.expected_doc_id !== null && row.expected_doc_id !== undefined) rows.push(row);
+  }
+  return rows;
+}
+
+const children = [];
+function killChildren() {
+  for (const child of children.splice(0).reverse()) {
+    try { child.kill(); } catch { /* gone */ }
+    if (IS_WIN && child.pid && child.exitCode === null) {
+      try { spawnSync('taskkill', ['/PID', String(child.pid), '/T', '/F'], { stdio: 'ignore' }); } catch { /* best effort */ }
+    }
+  }
+}
+
+function waitForPortFile(portFile, timeoutMs) {
+  const t0 = Date.now();
+  for (;;) {
+    if (fs.existsSync(portFile)) {
+      const port = Number.parseInt(fs.readFileSync(portFile, 'utf8').trim(), 10);
+      if (Number.isInteger(port) && port > 0) return port;
+    }
+    if (Date.now() - t0 > timeoutMs) throw new Error(`host did not write its port within ${timeoutMs}ms`);
+    Atomics.wait(new Int32Array(new SharedArrayBuffer(4)), 0, 0, 50);
+  }
+}
+
+async function fetchJson(url, opts) {
+  const res = await fetch(url, opts);
+  return { status: res.status, body: await res.json().catch(() => null) };
+}
+
+async function main() {
+  const args = parseArgs(process.argv.slice(2));
+
+  // Resolve weights the same way the host's resolvers do: explicit env first,
+  // then the repo's staged models/ dir. The embedder is REQUIRED (no staged
+  // weights means nothing to measure); the reranker is optional (row records
+  // reranker=none, retrieval degrades to RRF-only in the host under test).
+  const embedderDir =
+    process.env.TRAININGAPP_EMBEDDING_MODEL_DIR ?? path.join(repoRoot, 'models', 'bge-small-en-v1.5');
+  if (!fs.existsSync(path.join(embedderDir, 'onnx', 'model.onnx'))) {
+    throw new Error(`embedder weights not staged: ${embedderDir} (run scripts/prepare-models.mjs)`);
+  }
+  let rerankerDir = process.env.TRAININGAPP_RERANKER_MODEL_DIR ?? null;
+  if (rerankerDir === null) {
+    const candidate = path.join(repoRoot, 'models', 'ettin-reranker-32m-v1');
+    if (fs.existsSync(path.join(candidate, 'onnx', 'model.onnx'))) rerankerDir = candidate;
+  }
+
+  // The exact retrieval config the host under test will resolve (defaults or
+  // the operator's TRAININGAPP_RETRIEVAL_* overrides) — recorded in the row.
+  const configUrl = pathToFileURL(
+    path.join(repoRoot, 'desktop', 'dist', 'main', 'backend', 'retrieval', 'config.js'),
+  ).href;
+  const { resolveRetrievalConfig } = await import(configUrl);
+  const retrievalConfig = resolveRetrievalConfig(process.env);
+
+  const work = fs.mkdtempSync(path.join(process.env.TEMP ?? repoRoot, 'retrieval-bench-'));
+  try {
+    const token = `bench-${process.pid}-${Date.now()}`;
+    const portFile = path.join(work, 'port.txt');
+    const storePath = path.join(work, 'store.sqlite');
+    const env = { ...process.env };
+    env.TRAININGAPP_EMBEDDING_MODEL_DIR = embedderDir;
+    if (rerankerDir !== null) env.TRAININGAPP_RERANKER_MODEL_DIR = rerankerDir;
+    delete env.TRAININGAPP_DESKTOP_EMBEDDER;
+    delete env.TRAININGAPP_DESKTOP_ENGINE;
+    const host = spawn(
+      process.execPath,
+      [
+        path.join(repoRoot, 'desktop', 'dist', 'main', 'backend', 'dev-server.js'),
+        '--port-file', portFile,
+        '--mode', 'node',
+        '--token', token,
+        '--store-path', storePath,
+      ],
+      { cwd: repoRoot, env, stdio: ['pipe', 'inherit', 'inherit'] },
+    );
+    children.push(host);
+    const hostPort = waitForPortFile(portFile, 30000);
+    const headers = { 'content-type': 'application/json', 'x-desktop-token': token };
+    const base = `http://127.0.0.1:${hostPort}`;
+    const deadline = Date.now() + 30000;
+    for (;;) {
+      try {
+        const res = await fetch(`${base}/health`, { headers });
+        if (res.status === 200) break;
+      } catch { /* not accepting yet */ }
+      if (Date.now() > deadline) throw new Error('host did not become healthy within 30s');
+      await new Promise((r) => setTimeout(r, 100));
+    }
+
+    const ingest = await fetchJson(`${base}/ingest`, {
+      method: 'POST',
+      headers,
+      body: JSON.stringify({ directory: path.join(repoRoot, 'eval', 'corpus') }),
+    });
+    if (ingest.status !== 200 || !ingest.body || ingest.body.documents !== 8) {
+      throw new Error(`ingest failed status=${ingest.status} body=${JSON.stringify(ingest.body)}`);
+    }
+    console.log(
+      `[retrieval_bench_driver] ingested ${ingest.body.documents} docs / ${ingest.body.chunks_added} chunks`,
+    );
+
+    const corpusBasenames = new Set(
+      fs.readdirSync(path.join(repoRoot, 'eval', 'corpus')).filter((name) => name.endsWith('.md')),
+    );
+    const questions = loadInCorpusQuestions().slice(0, args.queries);
+    if (questions.length < args.queries) {
+      throw new Error(`only ${questions.length} in-corpus eval questions available, need ${args.queries}`);
+    }
+
+    const latencies = [];
+    for (const question of questions) {
+      const started = performance.now();
+      const res = await fetchJson(`${base}/search`, {
+        method: 'POST',
+        headers,
+        body: JSON.stringify({ query: question.question, n_results: 5 }),
+      });
+      const ms = performance.now() - started;
+      latencies.push(ms);
+      if (res.status !== 200 || !Array.isArray(res.body) || res.body.length < 1) {
+        throw new Error(`search failed query="${question.id}" status=${res.status}`);
+      }
+      for (const row of res.body) {
+        const source = typeof row.source === 'string' ? row.source.replace(/\\/g, '/').split('/').pop() : null;
+        if (!source || !corpusBasenames.has(source)) {
+          throw new Error(`non store-backed row query="${question.id}" source=${JSON.stringify(row.source)}`);
+        }
+      }
+    }
+
+    const p50 = percentile(latencies, 50);
+    const p95 = percentile(latencies, 95);
+    const max = Math.max(...latencies);
+    console.log(
+      `[retrieval_bench_driver] ${latencies.length} /search round trips; ` +
+        `p50=${p50.toFixed(1)}ms p95=${p95.toFixed(1)}ms max=${max.toFixed(1)}ms`,
+    );
+    const outcome = p95 <= P95_BUDGET_MS ? 'pass' : 'fail';
+    const row = {
+      surface: 'desktop-retrieval',
+      machine: process.env.BENCH_MACHINE_TAG ?? 'devstation',
+      embedder: path.basename(embedderDir),
+      reranker: rerankerDir === null ? 'none' : path.basename(rerankerDir),
+      topk: retrievalConfig.topK,
+      multiplier: retrievalConfig.candidateMultiplier,
+      p50_ms: Math.round(p50),
+      p95_ms: Math.round(p95),
+      max_ms: Math.round(max),
+      outcome,
+    };
+    const rendered = JSON.stringify(row, null, 2);
+    if (args.json) fs.writeFileSync(args.json, rendered);
+    console.log(rendered);
+    process.exitCode = outcome === 'pass' ? 0 : 1;
+  } finally {
+    killChildren();
+    try { fs.rmSync(work, { recursive: true, force: true }); } catch { /* best effort */ }
+  }
+}
+
+main().then(
+  () => {
+    // onnxruntime-node may keep the child's worker threads alive past kill;
+    // the child is force-killed above, so just make sure WE exit.
+    setTimeout(() => process.exit(process.exitCode ?? 0), 250).unref();
+  },
+  (err) => {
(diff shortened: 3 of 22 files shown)


zaxbysauce:
## Swarm PR Review — PR #102 (issue #65, B7 hybrid retrieval)

Full layered review: 6 base-dimension lanes + 7 risk-family micro-lanes (12 of 13 planned lanes settled; `compatibility-delivery` was unrecoverable after 3 dispatch attempts — see Coverage gap below) → 6 independent reviewer passes validating 46 raw candidates against source → critic challenge on the highest-severity findings. This comment reflects post-critic verdicts, not raw candidate claims.

### Verdict: **APPROVE_WITH_NOTES**

Nothing here blocks merge. Zero CRITICAL/HIGH findings survived independent reviewer + critic validation. The review did surface real, cheap-to-fix defects and one evidentiary-honesty issue worth addressing before or shortly after merge.

---

### Confirmed findings (post-critic)

**F-1 — MEDIUM — Reranker worker jobs can hang forever with no timeout or recovery**
`desktop/main/backend/retrieval/reranker.ts:119-160`
A pending `score()`/`embed()` job only settles via a matching `jobId` message, an `'error'` event, or an `'exit'` event **with a non-zero code**. Two source-confirmed gaps: (1) a clean worker exit (`code === 0`) never calls `failAll()`, leaking every pending job forever; (2) a malformed/unrecognized reply is silently dropped with no fallback settle path. No per-job timeout exists anywhere upstream (verified via repo-wide grep — the only timeouts in `desktop/main` are on the sidecar health probe and LLM generation, not retrieval). This defeats the PR's own stated isolation contract ("a broken reranker worker must never 500 `/search`," `hybrid.ts:226-228`) — a wedged worker is neither a clean degrade nor a 500, it's unhandled.
*Fix:* add a per-job timeout in `score()`/`embed()`, and call `failAll()` unconditionally in the `'exit'` handler regardless of exit code.

**F-2 — LOW — Reranker-failure degrade path misattributes embedder faults and wastes duplicate work**
`desktop/main/backend/retrieval/hybrid.ts:182-185,229-241`
The degrade `try/catch` wraps the entire `runSearch` call, including the initial `embedder.embed()`. In practice this can't manifest as a wrong *log message* against a healthy reranker (index.ts:166-176 ties the reranker and embedder-worker together, so "reranker down but embedder up" isn't a producible wiring) — but under any rerank-only fault, the retry re-runs the full pipeline (embed + vector query + FTS query) a second time outside any try/catch, wasting one duplicate embed + two DB round-trips per request. The related claim that this "undercuts the PR's p95 latency numbers" doesn't hold — the retry skips the (expensive) reranker entirely, and no fault-mode latency was ever measured in either direction, so there's no basis to assert a budget breach. The "reloads a ~127MB model every request" framing was also imprecise — the actual repeat-cost driver is `rerank-worker.ts:75-78` resetting `rerankPromise` on failure (retry-reload), not a worker respawn (`ensureWorker()` only respawns after a worker *exit*, which the current catch-everything design avoids).
*Fix:* move the try/catch inside `hybridRetrieve` around just the `options.reranker.score(...)` call and fall through to the already-computed fused ordering; delete the outer wrapper. One edit fixes both the misattribution and the duplicate work.

**F-3 — LOW — Document-surface detach is dead code after this PR's refactor**
`desktop/main/backend/index.ts:71,186,222-225,374-400`
Base revision assigned `this.surface = attachStoreSurface(...)`; this PR changed the helper to return `void` and dropped the assignment. `this.surface` is now permanently `null`, so `stop()`'s `if (this.surface !== null) engine.attachDocumentSurface(null)` guard never fires — the engine's document surface is never detached on shutdown. (The sibling `this.retrieval` field, added by this same PR, correctly follows the assign-then-null pattern this one now lacks — found independently by two review lanes.) Severity capped at LOW: `document-surface.ts`'s methods all re-check `getStore() === null` before doing anything, so the retained surface degrades safely rather than misbehaving; the actual retention is engine-side (`llama-engine.ts`), not a host-level leak. No test covers `attachDocumentSurface(null)` on shutdown.
*Fix:* restore `this.surface = attachStoreSurface(...)` at the `start()` call site (`index.ts:186`), or re-add a return value to the helper.

**F-4 — LOW — Relevance-floor env parser accepts partial-parse garbage**
`desktop/main/backend/retrieval/config.ts:67-73` (execution-verified: `Number.parseFloat('0.5garbage')` → `0.5`)
`floorOrFallback` uses bare `parseFloat` with no full-match pattern gate, unlike the sibling `positiveInt` (`/^\d+$/` full match). `TRAININGAPP_RETRIEVAL_RELEVANCE_FLOOR="0.5garbage"` silently yields `0.5` instead of falling back to the documented calibrated default (`0.569387`), contradicting both the code's own "defaults on invalid input" comment and the test suite's stated invalid-input coverage (which never includes a partial-parse case). Operator-controlled only, not attacker-reachable.
*Fix:* gate with a full-match numeric pattern before `parseFloat`, mirroring `positiveInt`.

**F-5 — LOW — `NaN` reranker scores silently pass the "invalid score" guard**
`desktop/main/backend/retrieval/hybrid.ts:190` (found while validating an unrelated candidate)
The fallback guard is `typeof score === 'number' ? score : 0`. `typeof NaN === 'number'` is `true`, so a `NaN` score (reachable if a reranker's logits array has an out-of-bounds/undefined entry) is never replaced with `0` — it then silently fails the `>= floor` filter and the chunk is dropped with no diagnostic distinguishing "genuinely irrelevant" from "malformed reranker output."
*Fix:* one token — `Number.isFinite(score) ? score : 0`.

**F-6 — LOW — No upper bound on `topK`/`candidateMultiplier`/`rrfK` config**
`desktop/main/backend/retrieval/config.ts:51-59,79-85` + `hybrid.ts:137,147,156,111`
`positiveInt` has no ceiling; a large operator-set `topK`/`candidateMultiplier` can push the fused candidate count past SQLite's bound-parameter limit inside `loadChunkRows`'s `IN (...)` clause. Mitigated to a clean 500 (`server.ts:591-593` catches everything) rather than a crash; operator-misconfiguration-only, not attacker-reachable (`n_results` from the request never feeds this).
*Fix:* clamp `positiveInt` to a documented ceiling.

**F-7 — LOW/advisory (evidentiary honesty) — No CI workflow exercises the real reranker scoring path**
`.github/workflows/desktop-build.yml` (all 3 checkouts lack `lfs: true`) + `desktop/main/backend/retrieval/reranker.ts:50,61-70,98-111`
No workflow stages ONNX reranker weights or sets `TRAININGAPP_RERANKER_MODEL_DIR`; `models/ettin-reranker-32m-v1` isn't tracked in the repo at all. `resolveRerankerModelDir()` therefore returns `null` in every CI run, and no test imports/exercises `rerank-worker.ts`'s real scoring path. **Correction to the initial framing:** this does *not* mean "zero coverage of the rerank path" as first stated — the no-reranker degrade path *is* tested (`b7-hybrid-retrieval.test.ts`), and the PR's own head-commit message already documents the narrower, accurate claim ("291/291 with staged weights AND 291/291 with an empty tree"). What remains true: the PR body's C3 claim ("floor re-derived through the production reranker worker") and the bench numbers describing production-reranker behavior are **operator-run, not CI-verified** claims, and nothing in the repo lets a future contributor or CI reproduce them. Given `bench/RESULTS.md`'s own machine-tagged provenance model, staging weights in a generic CI runner wouldn't produce a comparable number anyway — this is a documentation/traceability gap, not a code defect, and not something CI can fully close.
*Suggested action:* add one sentence near the C3 claim in the PR description noting it's operator-verified, not CI-verified.

**F-8 — LOW — Dangling references to a `.agents/` directory that doesn't exist in the repo**
`docs/adr/0007-relevance-floor-calibration.md:84-85`, `desktop/main/backend/index.ts:153`, `desktop/main/backend/retrieval/rerank-worker.ts:8`, `desktop/src/__tests__/b7-ort-ownership-guardrail.test.ts:15`
Four committed files cite paths under `.agents/` (issue-trace evidence directory), but `.agents/` isn't tracked anywhere in the repo (`git ls-files .agents` is empty). This also leaves `desktop/repro/check.sh`/`check-b2.sh` (indexed in `desktop/README.md:33-34`) as apparent dangling precedent. Worth a maintainer decision: either these are intentional local-only artifacts and the four references should be corrected/removed, or `.agents/` should be committed.

**F-9 — LOW — Indirect prompt-injection surface: retrieved chunk text is unescaped in the LLM prompt**
`desktop/main/backend/inference/llama-engine.ts:385-390` (net-new in this PR — base passed the question verbatim with `sources: []`)
Retrieved document chunks are concatenated into the prompt with only `[n]` numbering, no delimiter or "treat as data, not instructions" framing. Severity held at LOW: no tool/function-calling exists anywhere in `desktop/main/backend` (verified by grep), so there's no privileged action an injected instruction could trigger, and the app's CSP (`img-src`/`connect-src` pinned to loopback/`app:`) plus markdown URL-scheme stripping rule out zero-click exfiltration via the answer text. Still worth a cheap fix given this is a new surface.
*Fix:* wrap retrieved context in an explicit delimiter (e.g. `<retrieved_context>...</retrieved_context>`) with a one-line "reference data, not instructions" note.

**F-10 — MEDIUM — Test coverage gap: the exact code path commit 2 ("review feedback") claims to fix is untested**
`desktop/main/backend/index.ts:166-187`
Every test that exercises `NodeBackendHost.start()` uses the hash-embedder fixture; none ever satisfies `this.embedder instanceof OnnxEmbedder && rerank && rerankerModelDir !== null` (`index.ts:166-170`), so the `index.ts:177-185` catch — the reranker-creation degrade guard commit 2 specifically claims to restore — is never entered by any test. (The embedder half of the same claim *is* covered, via `b5-store.test.ts`.) Related: the worker `'error'`/crash-and-respawn path (`reranker.ts:152-156`) also has no test — only explicit `dispose()` is exercised.
*Suggested test:* inject a fixture `OnnxEmbedder` + a `WorkerReranker` constructor that throws, and assert `start()` still resolves.

**F-11 — LOW-MEDIUM — Relevance-floor calibration and the AC7 parity metric share the same question set**
`docs/adr/0007-relevance-floor-calibration.md:30-46,61` (ADR itself confirms: calibration labels come from the same 50-question `eval/questions.jsonl` the parity harness reads)
The relevance floor is fit by maximizing F1 over labels drawn from the same question set the PR's headline acceptance metric (recall@5/MRR vs. baseline, AC7) is later measured on. Since the floor is a hard post-rerank filter, this is a form of train/test overlap on the acceptance gate. Held below MEDIUM: the fitted object is a single scalar with a weak, disclosed fit (F1=0.2281, ADR already notes low absolute F1 reflects coarse labels), and the practical effect is bounded. The real gap is disclosure — ADR-0007 doesn't flag the overlap.
*Suggested action:* add one paragraph to ADR-0007 noting the shared question set, or re-run AC7 with the floor held out to show the acceptance metric isn't floor-sensitive.

**F-12 — LOW — Test cleanup ordering can leak a live worker thread on assertion failure**
`desktop/src/__tests__/b7-rerank-worker.test.ts:90-125`
`runner.dispose()` sits inside the `try` block rather than `finally`; if a prior assertion throws, the worker is never terminated, which can keep the vitest process alive and compound flakiness specifically when a test is already failing.
*Fix:* move `dispose()` into the existing `finally`.

---

### Advisory / non-blocking notes (confirmed, low-severity, not requiring action before merge)

- `sanitizeFtsQuery` (`hybrid.ts:88-100`) has no internal length cap or type guard — currently safe only because `server.ts` bounds and type-checks the query before it arrives. Defense-in-depth gap for any future direct caller.
- The worker `'error'` event handler doesn't null `this.worker` (only `'exit'` does) — a narrow timing window where a job could be posted to an already-dying worker; self-heals quickly via the subsequent exit event in the common case.
- `ensureWorker()` has no backoff/circuit-breaker on repeated crashes — each crash pays a full model reload before the next failure.
- The worker dispatches each incoming message as a non-awaited async IIFE, contradicting the module's own "serializes rerank jobs" doc comment — no correctness bug (response routing is `jobId`-keyed and safe), but doc-vs-code mismatch.
- `resolveEmbedder`'s explicit-env-dir fallback (pre-existing code, `embedder.ts:193-200`) bypasses the staged-weights validity check — an invalid `TRAININGAPP_EMBEDDING_MODEL_DIR` silently builds a broken embedder instead of failing fast at `start()`. Pre-existing, not introduced by this PR, but newly load-bearing given B7's `instanceof OnnxEmbedder` gate.
- `reranker.ts:50`'s `RERANKER_WEIGHTS_MIN_BYTES` 10MB threshold — initially flagged as an unexplained magic number, but the rationale is documented on the preceding line; no action needed.
- The RRF fusion formula, rank indexing, and dedup-by-chunk-id merge logic were independently checked by two lanes and found mathematically correct (formula matches the canonical form; reciprocal-rank contributions are summed, not clobbered, when a chunk appears in both legs).

### Disproved candidates (raised by explorer/reviewer lanes, disproved on independent validation)

- *"C8 p95=517ms claim is unverifiable/fabricated"* — disproved. `bench/RESULTS.md`'s own provenance rule only scopes to the committed table rows (which comply); the 517ms figure sits inside the driver's own recorded `max_ms=691` — this is measurement variance the PR body already labels distinctly, not a violation.
- *"C9 conformance 10/10 has no implementing artifact"* — disproved by one reviewer lane (found to map to the base-URL conformance suite; by-design detach-path behavior), though a second lane flagged the same gap as unresolved — worth a maintainer clarification but not a blocking finding either way.
- FTS5 query sanitizer: no injection/breakout found across two independent adversarial traces (operator-word, quote-breakout, column-filter, star/caret, whitespace-only) — the drop-whole-token strategy (not partial escaping) closes off every tested bypass class.
- No double-worker-spawn race: the theorized concurrent-`start()` hazard requires a caller this app doesn't have (single `whenReady()` invocation behind a single-instance lock).
- `rerank-worker.ts`'s score-array-shorter-than-window theory: disproved — the scoring loop always returns exactly `candidates.length` entries; see F-5 above for the adjacent real bug this candidate nearly found.
- Several LOW/info-level security and privacy candidates (config-floor-as-attack-surface, worker postMessage validation, log-content leakage, "once-per-lifetime" warning suppression) were examined and found to be non-exploitable or working-as-intended on independent review — full detail available on request.

### Coverage gap (disclosed per protocol)

The `compatibility-delivery` base-dimension lane (API/schema/config compatibility, docs consistency, packaging) could not be completed — 3 independent dispatch attempts all lost their findings to a session Stop-hook interaction after full investigation (each ran ~19-30 tool calls but returned only meta-commentary). Partial coverage of this dimension's concerns exists via the `api-schema-migrations` and `intent-architecture` lanes above (F-7, F-8, and the advisory notes). Flagging honestly rather than fabricating this dimension's findings.

### Merge recommendation

**APPROVE_WITH_NOTES.** No CRITICAL or HIGH findings survived independent validation. F-1 (worker hang) and F-10 (untested degrade path) are the two most worth addressing — ideally before merge, since both are small, well-scoped fixes with clear repro paths. F-2/F-3/F-4/F-5/F-6/F-9/F-12 are one-line-to-small fixes suitable for merge now or as an immediate follow-up. F-7/F-8/F-11 are documentation/traceability items for the maintainer's judgment.

---
*Generated by an automated multi-agent swarm review (12/13 explorer lanes → 6 independent reviewer passes → critic challenge on highest-severity findings). Findings above reflect post-critic verdicts; raw/pre-critic severities were higher for several items and were revised down on evidence.*

zaxbysauce:
**Swarm PR review complete** (run `pr102-review-20260910`): 6 base lanes + all 11 risk-family micro lanes over `e23b460`, independent cross-model validation (MiniMax-M3 shards), and a two-pass critic challenge on the HIGH/CRITICAL band.

**Validated: 17 findings → actionable subset fixed in efafa40** (this head):

| Finding | Severity (final band) | Outcome |
|---|---|---|
| PRR-004 reranker failure latch retried per query | MEDIUM | FIXED (latch persists for surface lifetime) |
| PRR-005 non-finite reranker scores poison sort | MEDIUM | FIXED (finite guard, map to 0) |
| PRR-010 query length unbounded at retrieval layer | MEDIUM | FIXED (8000-char cap, defense-in-depth) |
| PRR-011 unbounded env topK → absurd vec0 k | MEDIUM | FIXED (upper bounds + default fallback) |
| PRR-006 dispose(): terminate() rejection orphaned pending jobs | MEDIUM | FIXED (try/finally) |
| PRR-007 worker clean-exit left pending jobs unsettled | MEDIUM | FIXED (orphan-exit failAll, dispose-exempt) |
| PRR-012 bench driver had no per-query deadline | MEDIUM | FIXED (30s AbortSignal) |
| PRR-008/017/018 test gaps (degrade path, sanitize tokens, row ordering) | MEDIUM/LOW | FIXED (additive 6-test hardening spec) |
| PRR-001 no in-repo eval-parity regression guard | MEDIUM | FOLLOW-UP issue (bench-floors-style eval floor; CI structurally cannot run weighted eval — LFS pointers) |
| PRR-002/003 explicit degrade-path assertions | LOW | FOLLOW-UP (implicit coverage exists and is load-bearing — it caught the e23b460 regression) |
| PRR-009 WorkerEmbedder unit routing | MEDIUM | documented residual (C7/C8 are the runtime discriminators) |
| PRR-016/019/020 | LOW | dispositioned in the closure ledger (no change / plan-pinned / standard practice) |
| PRR-013/014/015/021 | — | REJECTED with evidence (pre-existing / empirically disproved / not defects) |

**Verification of the fixes**: fresh cross-model reviewer round — 297/297 desktop suite, frozen specs unperturbed (C1/C2/C4 re-run green), C8 end-to-end `C8 PASS p95=351ms`, tracer C7 parity re-run `recall@5=0.640 mrr=0.384` at this head. Full ledger: `.zcode/pr-review/pr102-review-20260910/` (candidate list, trigger-eval ledger, closure artifacts; per-lane raw outputs are summarized in the candidate list).

zaxbysauce:
Follow-up tracking issue for the deferred review findings (eval-parity floor, degrade-path assertion hardening): #103 — filed at merge per the standing agreement; everything else from the review was fixed in efafa40 or dispositioned in the closure ledger above.

PR #107, opened 2026-09-12
feat(packtool): D2 offline narration transcription — content-hash cache + ASR overlay (issue #78)

zaxbysauce:
Closes #78

## Summary

D2 of Workstream D: a build-machine-only offline transcription step that recovers narration text for every audio/video asset in a Storyline publish that has no native caption/transcript sidecar, caches results by content hash, and feeds D1's per-slide document schema via `transcript_source`/`transcript_text` — with `transcript_source: "asr"` distinguishable from `"sidecar"`.

- **`packtool/storyline/transcribe.py` (new)** — Python CLI: inventories a publish's audio/video media objects (data.js `assetLib` + fully recursive slide-payload walk; audio objects live in layer `audiolib[]`, videos in layer `objects[]`), skips media already covered by a native `<objectId>_transcripts.js` sidecar, transcribes the rest with faster-whisper (**`distil-large-v3`**, `int8` CPU, language `en` — the named choice; audio decode via the FFmpeg libraries bundled with faster-whisper, so MP3s and MP4 audio tracks take one path with **no system ffmpeg**), caches by content hash, and writes one sidecar-format transcript per media OBJECT id.
- **Content-hash cache (issue-mandated key)** — `composite = "<format_version>|<sha256(media bytes)>|<model>|<compute_type>"`, entries at `<cache-dir>/transcripts/<sha256[:12]>/<sha256(composite)>.json` storing cues + key components; the loader compares stored components to the request and treats ANY mismatch as a miss, so a stale or mis-keyed entry can never be served. Re-runs over unchanged media make **zero ASR calls** (spy-asserted).
- **Output** — `<asrDir>/<objectId>_transcripts.js` in the SAME byte format as native sidecars (`const data = {...}; window.globalLoadJsAsset(...)`), cues `{"start_ms","text"}` plus `"source":"asr"` provenance — so `packtool/storyline/video-refs.ts` consumes ASR output through its **unchanged** decode path.
- **D1 consumer (opt-in)** — `packtool storyline extract <dir> --out <dir> --asr-dir <dir>` additionally resolves AUDIO objects and sidecar-less media from the store (`transcript_source: "asr"`, `narration_ref: "<id>_transcripts.js"`, ASR-store-relative and deliberately not in `source_files`). `TranscriptSource` gains `'asr'` in both `types.ts` and `video-refs.ts`. Without `--asr-dir`, extraction output is byte-identical to before (golden-pinned).
[...]

Issue #82, opened 2026-09-05
[Workstream D] PR 6 of 7: Learn panel — answer + "where to learn this" with Open-in-training deep links (both backends)


zaxbysauce:
**Roadmap:** D6 · **Workstream D: Training pack and Learn panel**
**Sequence:** PR 6 of 7 (tracking slot; dependencies define execution order)
**Roadmap baseline:** 2026-09-05 · **Priority:** P0 · **Repo revision:** b2ddda0

## Outcome
When this closes, every `/ask` and `/ask/stream` response (both the Python backend and, once #81 lands, the Electron-hosted backend per ADR-0003) carries a `learn[]` array — the union of direct training-slide hits and cross-linked slides of the cited doc chunks (#80), ranked and deduped by slide — and the chat UI renders a Learn panel showing "Section > Slide title", a one-line snippet, and an "Open in training" button that calls into #81's `TrainingPlayer.jumpToSlide()`. Retrieval quality against this feature is measured by a Learn hit@3 metric with an initial target of >= 0.70. This does NOT certify the underlying retrieval/reranking quality itself (owned by #65/B7 and its own eval gate), and does NOT implement ask-about-this-slide context pinning (#83/D7) — only the answer-time Learn surface and its ranking.

## Evidence motivating this slot
- Repo evidence for the exact response shapes this issue extends, both verified by direct read (2026-09-06):
  - `web_ui/src/lib/api/types.ts` [types.ts:77-83](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/web_ui/src/lib/api/types.ts#L77-L83):
    ```
    export interface AskResponse {
      question: string;
      answer: string;
      sources: string[];
      context_length: number;
      inference_time: number;
    }
    ```
    No structured provenance field exists today beyond a flat `sources: string[]` — this issue adds `learn?: LearnResult[]` here.
  - `api_server.py` [api_server.py:235-243](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/api_server.py#L235-L243):
    ```python
    class QuestionResponse(BaseModel):
        """Response model for question answers."""

        question: str
        answer: str
        sources: List[str]
        context_length: int
        inference_time: float
    ```
    Same gap on the Python side — both models must gain the field in this issue for cross-backend parity, per the template's "both backends" requirement.
[...]

Issue #81, opened 2026-09-05
[Workstream D] PR 5 of 7: Embedded Storyline player service + slide-state events (per the A8 recipe)


zaxbysauce:
**Roadmap:** D5 · **Workstream D: Training pack and Learn panel**
**Sequence:** PR 5 of 7 (tracking slot; dependencies define execution order)
**Roadmap baseline:** 2026-09-05 · **Priority:** P0 · **Repo revision:** b2ddda0

## Outcome
When this closes, the OpMed Storyline pack is served same-origin under `app://training/<packId>/` inside Electron, a `TrainingPlayer` component embeds it as an iframe, and the app can programmatically jump to any of the 384 slides using the exact runtime call proven by #58 (A8)'s spike, emitting `slidechange` events with the current slide id/title polled from the player's own state. This does NOT implement the Learn panel UI that calls into this component (#82/D6) or ask-about-this-slide context pinning (#83/D7) — only the player service, component API, and navigation primitive.

## Evidence motivating this slot
- CONFIRMED `E:\ClaudeCode\OpMed CDP MicroLearning Companion_7-10-26\html5\lib\scripts\bootstrapper.min.js`: contains the literal strings `GetPlayer`, `GetVar`, `SetVar` (confirmed via direct grep, 2026-09-06). Per #58's critic correction (`critic_sol_verdict.md` line 28: "Presence does not prove a callable slide-jump method"), this issue must consume #58's PROVEN recipe, not re-derive it from string presence alone.
- CONFIRMED `html5/data/js/data.js` `playervars` array includes entries named `projectSlideNumber` and `projectSlideTitle` (grep-confirmed: `"name":"projectSlideNumber"`, `"name":"projectSlideTitle"`, plus `projectSlideReference` and `projectSlidesViewed`) — these are the exact variable names to poll via `GetVar` for current-slide state, per `implementation_plan.md` line 73 ("poll `GetVar('projectSlideTitle'/'projectSlideNumber')` ~1 s").
- CONFIRMED `story_content/user.js` (present, non-empty, included by `story.html` via script tag) already contains a working `GetPlayer()` call pattern: `window.InitUserScripts = function() { var player = GetPlayer(); var object = player.object; var once = player.once; ... }` — this is the authored hook point Storyline itself uses to reach the player object at runtime, and is the lowest-risk integration point for this issue's polling/navigation code rather than reverse-engineering the minified bootstrapper independently.
[...]

PR #108, opened 2026-09-13
feat(packtool): D3 build-storyline — installable Storyline pack with prebuilt training index (issue #79)

### LINKED ISSUE #79: [Workstream D] PR 3 of 7: Build an installable Storyline pack with a prebuilt training index (packtool build-storyline)
**Roadmap:** D3 · **Workstream D: Training pack and Learn panel**
**Sequence:** PR 3 of 7 (tracking slot; dependencies define execution order)

## Outcome
When this closes, `packtool build-storyline <publishDir>` composes D1's extracted slide documents and D2's transcripts into a single installable knowledge pack (`source_class: "training"`) with prebuilt embeddings and FTS index plus the raw publish folder as player assets, so installing the pack requires zero client-side re-embedding. This does NOT implement the doc-to-slide cross-link table (owned by #80/D4), does not embed or serve the player (owned by #81/D5), and does not implement the Learn panel UI (owned by #82/D6).

- CONFIRMED `E:\ClaudeCode\OpMed CDP MicroLearning Companion_7-10-26\meta.xml`: no `imsmanifest.xml`, `tincan.xml`, or `cmi5.xml` anywhere in the tree; `story.html` sets `window.globals` with `lmsPresent: false, tinCanPresent: false, cmi5Present: false` (quoted: `lmsPresent: false,` / `tinCanPresent: false,` / `cmi5Present: false,`) — this is a pure static Web-only publish with no server dependency, confirming it can be copied wholesale into a pack's player-assets directory without an LMS runtime.
- CONFIRMED `story.html`'s `parseParams` function only recognizes `deviceType` and `devicepreview` query params — no documented `?slide=` deep-link parameter exists, which is why player embedding/navigation is a separate spike-dependent issue (#81), not bundled here; this issue only needs to ship the folder as static assets, not solve navigation.
- Repo evidence: `contracts/pack.schema.json` does not exist yet in this repo (`ls E:\ClaudeCode\trainingapp\contracts` fails, confirmed 2026-09-06) — this is prospective infrastructure from #68 (C1) that this issue's pack must conform to once merged, not an existing format to reuse today.
- Plan citation: `implementation_plan.md` line 59: `packtool build-storyline <publishedCourseDir>` is the named CLI verb this issue implements; line 61-64 describes the training-pack composition (extractor + Whisper cache + prebuilt embeddings) this issue wires together.

- #77 (D1) supplies the slide-document extractor and schema. #78 (D2) supplies cached transcripts. #73 (C6, packtool build-docs/verify/diff CLI) supplies the packtool CLI scaffold, the embedding/FTS prebuild mechanism, and the `verify`/`diff` commands this issue's `build-storyline` subcommand must plug into rather than reimplement.
- This issue OWNS: composing D1+D2 output into a `source_class: "training"` pack, bundling the publish folder as player assets, and invoking C6's embedding/FTS prebuild over the slide documents. It does NOT own the pack schema itself (C1/#68) or the packtool CLI framework (C6/#73).
- Not parallel-safe with #73 (hard dependency on the CLI scaffold existing); parallel-safe with #80/#81 once this merges, since those are downstream.

## Required scope
- [ ] `packtool/storyline/build-storyline.ts` (or equivalent under the `packtool` CLI per #73's scaffold): CLI subcommand `packtool build-storyline <publishDir> --out <packFile>` that (1) invokes D1's extractor, (2) resolves D2's cached transcripts for slides with `transcript_source: "missing"` at extraction time, (3) copies the publish folder (`html5/`, `story.html`, `story_content/` minus already-processed sidecars if redundant) into `assets/player/` inside the pack, (4) emits `pack.json` with `source_class: "training"`, course metadata (title, duration, author from `meta.xml`), and `docs[]` entries per slide document.
- [ ] Prebuilt embeddings + FTS index over all 384 slide documents + the course-outline document, using the embedding model/dims selected by ADR-0001 (#55/A5) — reference "per ADR-0001" rather than hardcoding a model choice in this issue.
- [ ] Reproducibility: running `build-storyline` twice against identical inputs produces a byte-identical pack (or byte-identical modulo a timestamp field, explicitly excluded from the hash check).
- [ ] `packtool build-storyline --verify` integration with C6's `verify` command (schema conformance, sha256 manifest of bundled assets).
- [ ] Size budget note: record the built pack's size (publish folder + index) in the PR description; flag if it materially affects the 7 GB installer budget referenced in #84/E1.
- [ ] Out of scope: cross-pack doc-to-slide linking (#80), player embedding/serving (#81), signature/tamper hardening beyond what C6/C8 already provide (#75).

## Acceptance
| Case | Required proof |
| --- | --- |
| Pack builds successfully | `packtool build-storyline <OpMed publish dir>` produces a pack file that passes `packtool verify` (test file: `packtool/storyline/__tests__/build-storyline.test.ts` or integration script) |
| Pack contains all 384 slide docs + outline | Assert `docs[]` length and presence of the course-outline document in the built `pack.json` |
| Zero client-side re-embedding required | Integration test: install the pack against the SQLite store (#59/B5's schema, once available) and confirm no embedding model invocation occurs during install (mock/spy assertion) |
| Reproducible build | Two builds from identical input produce identical `pack.json` docs[] content (hash comparison excluding declared volatile fields) |
| Player assets present and unmodified | Bundled `story.html`/`html5/` byte-compare against source publish folder (spot-check a subset of files, not full corpus, to keep test fast) |

## Evidence and exit gate
Closing PR must attach: the built pack's `packtool verify` output, the reproducibility hash comparison, and the recorded pack size. Invalidated if #68's `pack.schema.json` changes after this pack is built (re-verify against the new schema) or if the embedding model ADR (#55) changes.

Treat PR 3 of 7 as a stable tracking slot, not merge order. The explicit dependencies govern execution; focused implementation issues supply evidence to later qualification issues without depending on those later issues closing. Re-anchor source and active-PR overlap before implementation. Close only the fully satisfied scope: wired production callers, meaningful focused validation and documented outcomes, with no silent deferral. Open one PR per issue titled with this issue's prefix; keep `contracts/` (API OpenAPI, store schema, pack schema) authoritative; add tests for every acceptance row; update docs touched by the change. Verification-only residuals may close with concrete current evidence rather than an unnecessary code patch.

| Slot | Issue | Work |
|---|---|---|

| Slot | Issue | Work |
|---|---|---|

| Slot | Issue | Work |
|---|---|---|

| Slot | Issue | Work |
|---|---|---|

| Slot | Issue | Work |
|---|---|---|

zaxbysauce:
Closes #79

# PR Description — packtool build-storyline (D3)

## Root Cause

Feature gap (roadmap slot D3, not a defect): at base 8212139 the repo had no way to produce an installable training pack. `packtool/cli.ts:19` rejected every non-`storyline extract` verb with the usage string and exit 2; `pack.json` had zero emitters anywhere in the repo (the only `source_class`-adjacent hit was the store schema's comment anticipating `'training'` at `contracts/store.schema.sql:33`); there was no verify surface, no embeddings/FTS prebuild, and no pack format. Both inputs the issue composes were already landed: D1's extractor (#77, PR #106) and D2's ASR transcript cache (#78, PR #107).

## Fix

Two new packtool verbs on the existing scaffold (`packtool/build/` modules, 6 new files), composing D1's per-slide documents + D2's cached transcripts + the raw publish folder into one installable zip pack with a prebuilt `index.sqlite`, plus `packtool verify`:

- `packtool build-storyline <publishDir> --out <pack.zip> [--asr-dir] [--embedder hash|onnx] [--embedding-model] [--id] [--version] [--name] [--published-at]` — extracts (D1, unchanged), emits `docs/` (slide docs byte-for-byte + the pack's outline document with title/duration/author from meta.xml), copies `html5/`, `story.html`, `story_content/` byte-for-byte to `assets/player/` (symlinks/junctions in the publish folder are refused, never dereferenced into the pack), chunks slide-aware (desktop text-chunker semantics, 256w/100ovl, goldens pinned against the desktop implementation), embeds (default: repo-staged bge-small-en-v1.5 ONNX via transformers.js, cls/normalize/fp32 — same conventions as the desktop runtime embedder, so prebuilt vectors are query-compatible by construction; `--embedder hash` is the deterministic CI fixture; the model is a build flag per the still-open ADR-0001 #55), and writes `index.sqlite` by applying `contracts/store.schema.sql` from disk (sqlite-vec 0.1.9 loaded before vec0 DDL; `meta.schema_version`/`embedding_model_id`/`embedding_dims` stamped; chunk ids = `sha256(docId:chunkIndex:normalized)` and vectors as JSON arrays — byte-matching `desktop/main/backend/ingest/pipeline.ts`). `pack.json` follows the #68 DRAFT manifest schema (fields marked draft-per-#68 pending C1; `source_class: "training"`); default pack id = `<title-slug>-<courseid>` (collision-safe, from meta.xml). Zip is deterministic: fixed per-entry dates from `--published-at` (including explicitly created directory entries — JSZip's auto-created folders otherwise stamp build time), sorted order, pack.json first. Two real-corpus builds with fixed inputs are byte-identical within a build environment (ONNX pinned to single-threaded sequential execution to keep float reduction order stable). Scope: byte-identity is guaranteed within one machine/onnxruntime build; the cross-environment guarantee applies to the `hash` embedder, which is fully deterministic everywhere.
- `packtool verify <pack.zip|packDir>` — manifest conformance against the #68 draft shape, per-doc sha256 re-hash (tamper detection), index stamp conformance (schema_version, dims, model id, row-count parity, docs-table hash set, packs-row id), the `assets/player/story.html` anchor, and path-safety (`assertSafeDocPath`) on every manifest-supplied path (`docs[].path` AND `index.path` — the latter added after the final critic demonstrated an unguarded read outside the pack root).
- `storyline/spine.ts` gains additive exports only (`readMetaElementAttribute`/`readMetaAuthor`/`readMetaProjectAttribute`); D1's `OutlineDoc` type, `extract.ts`, and the extractor's output bytes are untouched (C8 suite + goldens prove byte-stability).
- `.gitignore`: `!packtool/build/` (the new directory is SOURCE; the generic `build/` pattern is for output).

Out of scope, unchanged: doc-to-slide links (#80), player embedding/serving (#81), Learn panel (#82), pack schema authorship (#68 — this pack conforms to its draft and re-plugs when it lands), `build-docs`/`diff` CLI (#73 — `verify` is the minimal shared verb this issue needed; #73 can extend it), signatures (#75).

- Defect class: "identity/hash math, pack-path handling, and store DDL must each have exactly one shared, tested surface" — a second divergent sha256 identity implementation, an unguarded manifest-supplied path, or a copied DDL silently corrupts pack identity, enables traversal, or drifts the index from the frozen schema.
- Sweep result: 7 predicate hits, all dispositioned in `08a-recurrence-sweep.md` — including one real gap the final critic caught (unguarded `index.path`; fixed with a regression test) plus the swarm-pr-review round: player-asset symlink/junction refusal, parse-time `--published-at`/`--version` validation, unknown-manifest-field rejection, verify parity checks, atomic `--out` publish, and sqlite handle-leak hardening and one parity gap found by the sweep itself (missing DDL-substitution residue check; added).
- Guardrail: `assertSafeDocPath` is the single path-safety gate consumed by both compose and verify at two layers each, unit-tested against traversal/absolute/backslash/empty payloads. Demonstrated to bite by mutation probe (both layers neutered → the traversal negative test FAILS 1/6; restored → 6/6) and by the C1 tamper check (verify rejects a mutated bundled doc).

## Tests

- Regression test: `npm --prefix packtool test` -> 17 files / 62 tests PASS (20 tests from the original implementation + review-hardening tests: index-path traversal refusal, zero-docs manifest, malformed dims type, unknown manifest field, vec0 distance bound, stored-vector == embedder recomputation)
- Impacted suite: the same command pins the pre-existing D1/D2 suite (9 files / 38 tests at base, goldens byte-for-byte) -> PASS
- Build: `npm --prefix packtool run build` (tsc) -> clean
- Frozen acceptance replay: 11/11 checks PASS via repro-check at HEAD (base legs all show the recorded RED sentinels; logs in the trace)
- Real corpus (OpMed, 384 slides, real ONNX embedder, #78 ASR store): `packtool build-storyline ... && packtool verify ...` -> `verify: OK (docs=385)`; two identical builds -> byte-identical zips within the build environment (see reproducibility scope note below)
- Deferred-work scan: scan-deferred.sh -> clean

## Regression Protection

- `packtool/storyline/__tests__/build-storyline.ac{2,3,4,5,7,8}.test.ts` — one file per acceptance check (docs count incl. outline, install-without-embedding spy, reproducibility, player-asset byte-identity, transcript wiring, index conformance + vec0/FTS5 retrieval)
- `packtool/storyline/__tests__/build-chunker.test.ts` — chunker parity goldens derived from the desktop implementation + identity-formula pins (drift alarm if the desktop chunker changes semantics)
- `packtool/storyline/__tests__/build-storyline.negative.test.ts` — path traversal (docs[] + index.path), missing pack.json, tamper re-hash, onnx-without-weights loud failure, defaultPackId collision rule
- CI: desktop-build.yml already runs the full packtool suite on every PR touching `packtool/**` (hermetic — no LFS weights needed)

|---|---|
| AC1 pack builds and passes verify; tampered pack fails | Frozen check C1 PASS (base RED → head GREEN); `node packtool/dist/cli.js verify <pack>` → `verify: OK (docs=4)` / tampered → `sha256 mismatch`, exit 1 |
| AC2 docs[] = every slide doc + outline | Frozen check C2 PASS: fixture docs[] == 4 (3 slides + outline); real corpus docs[] == 385 (384 + outline) |
| AC3 zero client-side re-embedding | Frozen check C3 PASS: `installPackRows` copies rows into a store.schema.sql-conformant DB with embedder call-count spy == 0 |
| AC4 reproducible build | Frozen check C4 PASS: pack.json minus published_at hash-identical across builds + docs[] deep-equal + zips byte-identical; real-corpus double-build byte-identical within the build environment (scope note above) |
| AC5 player assets present and unmodified | Frozen check C5 PASS: story.html + 3 html5/ files byte-equal to source under assets/player/ |
| AC6 pack.json contract (source_class training, embedding/chunking blocks) | Frozen check C11 PASS; stamps: `source_class: "training"`, `embedding {model_id, dims, normalize}`, `chunking {slide-aware, 256, 100}` |
| AC7 D2 cache resolved into pack | Frozen check C6 PASS: sidecar + asr transcript text in pack docs AND index chunks; narration-less slide builds |
| AC8 index conforms and retrieves | Frozen check C7 PASS: meta.schema_version=1, dims match, row counts equal, vec0 MATCH self-hit, FTS5 MATCH hit |
| AC9 pack size recorded in PR | Real-corpus build: `opmed pack size: 196600575 bytes` (~187.5 MB — ~2.7% of the 7 GB #84 installer budget; C9 head log) |
| Docs | Frozen check C10 PASS: packtool/README.md documents build-storyline (usage, `--embedder hash or onnx`), `packtool verify` |

## Invariant Audit

No invariant/architecture-contract doc exists in this repo ("none documented"). Contract-level invariants that DO exist were audited:

- `contracts/store.schema.sql` (frozen v1, sqlite-vec 0.1.9 pin): not modified; index.sqlite applies it from disk with the dims token substituted — not touched (git diff empty on contracts/)
- `contracts/api.openapi.yaml`: not touched
- D1 extractor output bytes (goldens `tests/fixtures/storyline-mini/expected/`): not touched — C8 suite green; `git diff origin/master...HEAD -- packtool/storyline/extract.ts packtool/storyline/types.ts` is empty (spine.ts gains additive exports only)
- Desktop runtime conventions (embedder/chunker/pipeline row shapes): mirrored, not modified (desktop/ untouched)

## Risk and Rollback

- Risk level: low — all changes are inside `packtool/` (+ .gitignore negation); no runtime code (desktop/, web_ui/, root Python) is touched; CI hermeticity preserved (hash embedder; the transformers.js dependency is lazy-loaded and never imported by tests)
- Rollback: revert the PR (two commits); no migrations, no runtime coupling
- Residual risk: `pack.json` conforms to #68's DRAFT schema; when #68 lands with `contracts/pack.schema.json`, the pack should be re-verified against the real schema (re-plug is additive; flagged in README + pack-json.ts comments). ADR-0001 (#55) may re-pin the embedding model — a build-flag change (`--embedding-model`), stamped per build into pack.json + index meta, not a schema change.

## Waivers (or none)

none

## Review Closure (swarm-pr-review + swarm-pr-feedback)

A full swarm-pr-review pass (6 base lanes, 8 attested micro risk families, 3 independent reviewer shards, 1 critic pass) validated ~35 candidates; the swarm-pr-feedback round resolved them:

**Fixed in this PR (review round):** player-asset symlink/junction refusal (critic-confirmed arbitrary-file-inclusion channel — a junction in the publish folder was silently dereferenced into the distributed pack; now refused loudly, empirically re-proven); parse-time `--published-at` and `--version` validation (garbage previously flowed into pack.json/zip entries or produced packs that only failed at verify); unknown manifest fields rejected (draft-#68 additionalProperties alignment); verify cross-checks `index.schema_version` and full packs-row parity; docs.source_class flows from the manifest; transaction ROLLBACK failures no longer mask original errors; atomic `--out` publish + stale-artifact removal; verify sqlite-handle close hardening; better-sqlite3 handle-leak hardening in openStoreWithSchema/installPackRows; `PACKTOOL_AC` selector now routes acN to both test suites (no silent D1-only selection); vec0 self-hit distance bound; stored-vector == embedder-recomputation assertion; new negative tests (zero-docs manifest, dims-as-string, unknown field).

**Explicitly dispositioned (not fixed, with reasons):**
- Zip size/entry-count caps in verify (resource exhaustion = verifier availability only, build-machine tool): deferred to C8/#75 install-hardening scope.
- `@huggingface/transformers` postinstall weight in CI + onnx path not exercised in CI: accepted — CI is hermetic by design (no LFS weights); the onnx path is validated on the build machine (real 385-doc build, verify OK, determinism evidence attached).
- Corrupted-but-large ONNX weights gate: accepted (size gate documented; real weights validated on build machine).
- ac3 embedder spy is decorative: accepted — the zero-re-embedding property holds by construction (install is a pure row copy with no embedder parameter); row-count parity is still asserted.
- Pack-size figure is one build's record (AC9 asks to record; runs vary by a few bytes with non-fixed `--published_at`).
- "Query-compatible by construction": build-time vectors are float-identical to the documented runtime configuration; single-thread pinning means cross-environment bit-identity is not claimed for ONNX (hash embedder is exact).
- Latent/INFO items (future-#82 outline consumer, exotic-path zip/dir verify parity, multi-media-object fixture coverage, usage-line stderr noise, pre-existing decode.ts bounds): documented in the review handoff (`.zcode/pr-review/prr-108-20260913/feedback-handoff.md`), no repo change warranted in this PR.

## Merge status

PR head: 0c376f99e632fd81d519428d772e9b86a88e9180

## Issue Closure

Closes #79

## Trace

Full issue-tracer evidence (intake → red checkpoint → approved plan → implementation → recurrence sweep → independent review → final critic) recorded in `.agents/issue-traces/79-build-storyline-training-pack/` (local, git-excluded). Key identities: plan-critic APPROVE, implementation-review APPROVE (incl. delta round), final-critic APPROVE (round 2), all bound to `119daffa23fa68eb38dc3a43451907b9c171afa8`; after the swarm-pr-review + swarm-pr-feedback round, the review-fix commit `18bdc023aed6628c522bdfacc711e74b141a8b50` carries fresh APPROVE verdicts from the PR-review reviewer council, the PR-review critic, and the feedback reviewer + final critic (tree `ec8960d5a974cb0bbcf021b3850cc2f6e17919ea`).


zaxbysauce:
## Swarm review — PR #108 (packtool build-storyline, D3)

Reviewed at head `119daff` (base `8212139`). Profile B (Claude Code native subagents, no swarm controller). Depth tier **L** (2,665+ non-lockfile insertions, multi-subsystem, security-sensitive path-safety surface). Six base-dimension lanes + seven matched risk-family micro-lanes → independent opus reviewer validation (4 chunks, several with live execution probes) → opus critic challenge on all MEDIUM-severity confirmed items.

### Verdict: **APPROVE_WITH_NOTES**

No CRITICAL or HIGH findings survived reviewer+critic validation. Everything confirmed landed at LOW after critic scrutiny — mostly hardening/diagnosability gaps and two evidence-accuracy issues in the PR's own safety/test-provenance claims, not functional or security defects. The core feature works as described: `assertSafeDocPath` genuinely blocks path traversal, the DDL is read from disk (not duplicated), sqlite-vec loads before vec0 DDL, and the deterministic-zip mechanism is real (verified by direct code read).

### Obligation mapping (issue #79)

| Obligation | Status | Evidence |
|---|---|---|
| `build-storyline` composes D1+D2 into an installable pack with prebuilt index | MET | `packtool/build/compose.ts`, `index-writer.ts`; real 384-slide corpus build verified per PR body |
| `packtool verify` conformance/tamper check | MET | `packtool/build/verify.ts`; traversal and tamper paths execution-tested |
| pack.json conforms to #68 draft schema | MET (as draft) | `pack-json.ts` fields marked draft-per-#68 |
| Byte-identical reproducible builds | UNVERIFIABLE (not MET/NOT_MET) | Mechanism is real (verified fixed dates, `createFolders:false`, ONNX single-thread pin) but the specific quoted real-corpus sha256 could not be independently reproduced in this review; **the ONNX determinism pin itself has zero automated test coverage** — every test uses the `hash` embedder, never a real ONNX load |
| Closes #79 | Issue #79 is open and the diff addresses its stated acceptance scope; no conflicting closing PR found |

`.gitignore`'s `!packtool/build/` addition is present and correctly scoped (verified via `git check-ignore`).

### Confirmed findings (all LOW after critic review — none block merge)

1. **Dead code + misleading test comment on the "second" path-safety guard.** `verify.ts`'s `index.path` check (~L152-157), added in this PR's final commit as defense-in-depth, is unreachable through the public `verifyPack()` entry point — `validatePackManifest` (called first) already catches unsafe `index.path` and returns early. Proven by live execution (calling `verifyPack` with a traversal payload returns exactly one problem, in layer-1's message format). **No security gap** — layer 1 genuinely blocks traversal and is tested — but the regression test's inline comment ("both guard layers must refuse") is factually inaccurate, and layer 2 is dead. Fix: delete `verify.ts:152-157` and correct the comment, or make layer 2 independently reachable/tested.

2. **PR's "11/11 frozen acceptance checks" claim is not reproducible from the committed tree — and the existing driver produces a false green.** No `repro/` directory exists under `packtool`, and `vitest.config.ts`'s `PACKTOOL_AC` selector hardcodes `extract.${ac}.test.ts` (issue #77/D1) — it never routes to this PR's own `build-storyline.${ac}.test.ts` (issue #79/D3) files. The tests do run fine under plain `npm test` (confirmed: 7 files/13 tests green), so nothing is untested — but **`packtool/scripts/acceptance-check.sh AC2` through `AC8` currently print `PASS` for D1's tests, not D3's**, which is a real false-green hazard for anyone using that script to validate this issue's acceptance criteria going forward. Recommend extending the selector to cover `build-storyline.${ac}` and fixing the six stale header comments.

3. **Offline embedder doesn't set `env.allowRemoteModels=false`.** `embedder.ts`'s `OnnxEmbedder` never configures transformers.js's `env` before calling `pipeline()`. Narrower than first appears: the default and documented invocation always produces an **absolute** model path, which fails transformers.js's HF-repo-id regex and resolves purely locally (throwing loud on a missing file, no network call). Egress is only reachable if an operator hand-types a relative, HF-repo-id-shaped `--embedding-model` value, and even then it 404s against a nonexistent HF repo rather than silently swapping in a wrong model. Still worth two lines of hardening (`env.allowRemoteModels = false; env.allowLocalModels = true;`) to make the "offline" claim airtight; the same gap pre-exists, unguarded, in `desktop/main/backend/ingest/embedder.ts`.

4. **`packtool verify` has no decompression/entry-count ceiling.** Reading an untrusted `.zip` (verify's actual design contract — it's meant to detect tampering on packs the operator didn't build) has no size cap on the whole-zip read, per-entry decompression, or `docs[]` array length; `verify.ts`'s `materialize()` also writes the decompressed index unboundedly to `os.tmpdir()` (disk, not just heap). Capped at LOW since this is a one-shot local CLI process, not a service — failure is a self-inflicted, fail-closed OOM/disk-fill. Recommend a sane per-entry/cumulative size budget in both the read path and `materialize()`.

5. **Native sqlite handle leaks on schema-apply failure, in two places.** `index-writer.ts`'s `openStoreWithSchema` (~L70-75) and `installPackRows` (~L162-163) both open a `better-sqlite3` handle and call `sqliteVec.load()`/`db.exec(schema)` **outside** any try/catch — if either throws, the handle is never closed. Empirically reproduced on Windows: an open handle makes the caller's cleanup (`rmSync(scratch, {force:true})`) fail with `EPERM` (force does not suppress lock errors), which can mask the real error. Confirmed blast-radius is temp-scratch-only today (never a persistent path, dies at process exit, no shipped artifact is ever corrupted) — a diagnosability bug, not a durability one. `verify.ts` already has the correct `let db=null; finally { db?.close() }` pattern elsewhere in the same file; recommend applying it to both leak sites.

### Additional confirmed LOW findings (advisory, not critic-routed — none change the verdict)

- Explicit `--id`/`--version` CLI flags bypass `PACK_ID_PATTERN`/`SEMVER_PATTERN` validation at build time (only checked later by `verify`) — late, confusing failure rather than a security issue (id/version never reach a filesystem path).
- `verify.ts` checks the index schema version against a hardcoded `'1'` literal instead of the exported `STORE_SCHEMA_VERSION` constant; a 4th independent copy of "1" also lives in `contracts/store.schema.sql`'s seed row and in `desktop/main/backend/store/migrate.ts`. None currently disagree, but a future schema bump needs coordinated edits across all four.
- The manifest's `index.schema_version`/`sqlite_vec_version` fields are structurally validated but never cross-checked against the index's actual stamped values — a manifest could claim the wrong version and still pass `verify`.
- `compose.ts`'s doc comment claims `assertSafeDocPath` guards both compose and verify; compose actually never calls it (safe today only because compose's paths are program-controlled, not because it's guarded).
- No streaming path for zip assembly — `compose.ts` reads every staged file (including full player-asset bundles) fully into memory before writing; fine for the tested 384-slide corpus, untested for a pack with large media assets.
- `writeFileSync(outPath, ...)` on `--out` is a non-atomic in-place overwrite; a crash mid-write on a re-run could destroy a previously-good `pack.zip`.
- One test fixture (`build-storyline.negative.test.ts:46`) leaks an unregistered temp dir (found independently by 3 lanes).
- `cli.ts` returns exit code 1 for both "pack failed verification" and "packtool crashed" — distinguishable only by stderr prefix, undocumented.
- `pack.json`'s `embedding.model_id` is just the model directory's basename, not a content hash — a silent weight swap with the same directory name would be undetectable across builds (pre-existing repo convention, not introduced by this PR).

### Disproved / no-issue

- Explorer speculation about an "empty-pack passes verify as OK" scenario is unreachable — the outline document always contributes at least one non-empty chunk (execution-verified).
- The claimed reachability gap in `meta.xml` validation (bare `readFileSync` with no guard) is dead code — D1's earlier extraction step already reads and validates `meta.xml` first.

### Coverage disclosure

The `correctness-state` base-review dimension and the `test-infrastructure` risk-family micro-lane both hit a known harness issue in this repo (subagent final output occasionally swallowed by a Stop-hook interaction) and required one-to-two retries each before landing usable output; all retries eventually succeeded and are reflected above. No dimension or risk family was left uncovered.

### Suppressed as noise (disclosed, not dropped)

A handful of LOW-confidence/purely-cosmetic candidates were not independently reviewer-validated given time budget: test-file co-location convention (storyline/__tests__ vs a hypothetical build/__tests__), a latent tsconfig-exclude gap with no current effect, a benign transitive `onnxruntime-common` version split in the lockfile, and an unconfirmed guess about missing AC1/AC6 test files. None looked likely to survive review as actionable.

---

zaxbysauce:
## Correction to the review above

A second, independent critic pass (this repo's protocol routes MEDIUM security/evidence-integrity findings through critic review, and I ran two independent passes on this batch) caught a real error in finding #2 above and sharpened finding #3. Posting the correction rather than leaving it stand:

**Finding #2 was based on a mechanism misidentification — retracted as written.** I conflated two unrelated verification systems. `packtool/scripts/acceptance-check.sh` is explicitly `# Acceptance-check driver for issue #77 (Storyline extractor)` — it only accepts `AC1`-`AC8` and prints `CHECK ACn: PASS/FAIL`. The PR body's "11/11 frozen checks C1-C11 PASS via repro-check" claim uses a different, C-numbered system entirely, and the PR body itself says that evidence is recorded in `.agents/issue-traces/79-build-storyline-training-pack/` — explicitly **local and git-excluded by design**, not something that should appear in the diff. So this claim is genuinely **unverifiable from the committed tree**, but not because any selector is "unwired" — that causal chain doesn't hold, since the AC-selector and the C-numbered frozen checks were never the same mechanism.

What *does* survive, as a much smaller, separate LOW nit: `vitest.config.ts`'s `PACKTOOL_AC` selector still hardcodes `extract.${ac}.test.ts`, so if `acceptance-check.sh` (issue #77's driver) is ever mistakenly reused against this PR's `build-storyline.acN.test.ts` files by name-similarity, it would silently run the wrong (D1) tests. This is a real but minor footgun, not evidence the D3 acceptance criteria are unverified — the actual PR claim (`npm --prefix packtool test` → 17 files/59 tests) is independently confirmed passing.

**Finding #3 (offline embedder egress) — severity held at MEDIUM, not downgraded to LOW.** Both critic passes agree the mechanism is real and narrower than the initial trace suggested (an absolute model-dir path — the default — is already fail-closed with no network access; only a hand-typed relative, HF-repo-id-shaped `--embedding-model` value reaches the network branch). Where they disagree is severity: this codebase states "offline"/"air-gapped" as a product contract in multiple places (`packtool/README.md`, root `README.md`, `docs/security_hardening_guide.md`), not merely a CI convenience, so a hermeticity breach is in-scope harm even when the trigger requires an atypical invocation. Recommend keeping it MEDIUM and making `env.allowRemoteModels = false` (plus `env.allowLocalModels = true`) in `embedder.ts` a required fix rather than optional hardening — it also converts a confusing 404 into a clear "not found locally" error.

Net effect on the verdict: unchanged (**APPROVE_WITH_NOTES**). All other findings and the overall assessment above stand.

---

Issue #73, opened 2026-09-05
[Workstream C] PR 6 of 9: packtool CLI — build-docs, verify, diff, with prebuilt embeddings and FTS on a build machine


zaxbysauce:
**Roadmap:** C6 · **Workstream C: Knowledge packs and freshness**
**Sequence:** PR 6 of 9 (tracking slot; dependencies define execution order)
**Roadmap baseline:** 2026-09-05 · **Priority:** P1 · **Repo revision:** b2ddda06e86b63f81bdbedce220a3a2357bc9852

## Outcome
When this closes, a build machine (CI or a maintainer's workstation) can run `packtool build-docs` to turn a folder of source documents into a fully conformant Knowledge Pack (C1 schema) with prebuilt embeddings and FTS already computed, so that end-user installs never re-embed content client-side. This issue does NOT extract Storyline slides (that is D1/D3's `packtool build-storyline` command, a sibling CLI verb sharing this tool's plumbing) and does NOT implement install-side logic (C2/C3 consume this issue's output).

## Evidence motivating this slot
- CONFIRMED: `scripts/export_seed_chunks.py` (155 lines, orphaned per C1's evidence) is the closest existing prior art for "export a built index to a shippable format" — this issue effectively replaces and generalizes it rather than resurrecting it, since it targeted ChromaDB's internal collection format, not the pack.json schema this workstream defines.
- CONFIRMED: [vector_store.py:716](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/vector_store.py#L716) `add_chunks_with_embeddings` is the exact consumer-side entry point that makes "prebuilt embeddings, zero client-side re-embedding" achievable for the Python backend — this issue's packtool output must populate `index.sqlite`/an equivalent embeddings dump in a shape that entry point (or C3's Node twin) can load directly.
- CONFIRMED (audit_report.md 1.2): bge-small-en-v1.5 (desktop) vs snowflake-arctic-embed-m-v1.5 (web_ui) are DIFFERENT embedding models today — packtool's build must stamp the pack's `embedding.model_id`/`dims` (per C1 schema) so C8's hardening issue can refuse a pack built for the wrong model rather than silently serving garbage vectors.

## Prerequisites and ownership
[...]

Issue #70, opened 2026-09-05
[Workstream C] PR 3 of 9: Node PackManager lifecycle on the SQLite store, sharing C2's semantics and test vectors


zaxbysauce:
**Roadmap:** C3 · **Workstream C: Knowledge packs and freshness**
**Sequence:** PR 3 of 9 (tracking slot; dependencies define execution order)
**Roadmap baseline:** 2026-09-05 · **Priority:** P0 · **Repo revision:** b2ddda06e86b63f81bdbedce220a3a2357bc9852

## Outcome
When this closes, whichever native desktop backend ADR-0003 (A7) selects can install, supersede, roll back, and remove Knowledge Packs against the shared SQLite schema (B5) with byte-identical chunk-id semantics to the Python implementation (C2), proven by running the same fixtures through both and diffing chunk id sets. This issue does NOT re-derive the pack format (C1) or the chunk-id formula (C1/C2) — it implements the same formula against a different store backend.

## Evidence motivating this slot
- CONFIRMED: [document_processor.py:349](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/document_processor.py#L349) and [vector_store.py:603-605](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/vector_store.py#L603-L605) establish that the Python path-hash chunk-id scheme this issue must NOT copy; C1's content-hash formula is authoritative for both backends so a pack installed once (e.g. by packtool, C6) produces the same chunk ids whether later read by Python or Node — this is what makes cross-backend fixture parity testable at all.
- CONFIRMED: [vector_store.py:954-1047](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/vector_store.py#L954-L1047) `delete_document` is the reference implementation this issue's `deleteDocument` must match in observable behavior (delete from vector table + FTS/BM25-equivalent + registry), even though the underlying store (SQLite/sqlite-vec vs Chroma) differs.
- CONFIRMED (issues_draft.md dependency graph, authoritative): "C3 ... needs B5-B7" — this issue cannot merge before B5 (sqlite-vec interop + frozen `contracts/store.schema.sql`) exists, since PackManager writes rows into that exact schema.

## Prerequisites and ownership
[...]

Issue #54, opened 2026-09-05
[Workstream A] PR 4 of 8: Build the tier-0 eval set and a backend-agnostic retrieval/answer eval harness


zaxbysauce:
**Roadmap:** A4 · **Workstream A: Contract, evidence and decisions**
**Sequence:** PR 4 of 8 (tracking slot; dependencies define execution order)
**Roadmap baseline:** 2026-09-05 · **Priority:** P1 · **Repo revision:** b2ddda06e86b63f81bdbedce220a3a2357bc9852

## Outcome
When this closes, there is one curated question set and one runner that can score recall@k, MRR, abstain accuracy, and latency against ANY backend implementing the frozen API contract (#51). Every later quality decision — embedding/reranker bake-off (A5/#55), hybrid retrieval parity in Electron (WS-B PR 7/#65), Learn-panel hit-rate (WS-D PR 6/#82) — is measured against this same set, so numbers are comparable across surfaces and time. This issue does NOT itself pick a winning model or reranker, and does NOT certify answer quality on the shipped product — it only builds the measuring instrument.

## Evidence motivating this slot
- CONFIRMED no eval baseline exists today: `web_ui/scripts/eval/README.md` baseline section is `<fill in>` (per audit_report.md line 91 and the critic's corroborating citation at `:60-63`).
- CONFIRMED the eval harness that does exist does not exercise the full weighted retrieval pipeline: audit_report.md §2.3 states "Eval harness (`scripts/eval/`) does not exercise the weighted pipeline; baseline numbers unfilled."
- CONFIRMED training-slide provenance exists to build "expected training slide" fields once the Storyline extractor lands: `frame.js`'s `navData.outline.links[].displaytext` gives section titles (e.g. `"displaytext":"Launch Menu"` — CONFIRMED by direct grep of `html5/data/js/frame.js` in the OpMed package) and `data.js` gives `slideCount: 384` (CONFIRMED via `grep -o '"slideCount":[0-9]*' data.js` → `"slideCount":384`).
- CONFIRMED the relevance floor this eval set must validate is currently unrevalidated: `rag-orchestrator.ts:148-163` documents `MIN_CROSS_SCORE=0.2` as "not yet re-validated" for the ettin reranker (per audit_report.md line 78, corroborated by the critic's citation of `rag-orchestrator.ts:153-156`).

## Prerequisites and ownership
[...]

PR #109, opened 2026-09-13
feat(store,packtool): D4 doc-to-slide links — schema v2, compute kernel + links CLI, transactional recompute (issue #80)

### LINKED ISSUE #80: [Workstream D] PR 4 of 7: Recompute doc-to-slide links whenever active packs change
**Roadmap:** D4 · **Workstream D: Training pack and Learn panel**
**Sequence:** PR 4 of 7 (tracking slot; dependencies define execution order)
**Roadmap baseline:** 2026-09-05 · **Priority:** P1 · **Repo revision:** b2ddda0

## Outcome
When this closes, every document chunk from every active pack has its top-3 nearest training slides recorded in a `links` table, computed at pack-build time for shipped doc packs and recomputed client-side whenever a pack is installed or superseded — so a citation to a vendor PDF can also point at the exact training slide covering the same content, even when the user's question wording matches the manual but not the terse on-screen slide text. This does NOT implement the Learn panel UI (#82/D6) or the player itself (#81/D5) — only the link computation and storage.

- Plan citation: `implementation_plan.md` line 66: "Doc <-> slide cross-link table (built at pack time): embed every slide doc and every doc chunk; for each doc chunk store its top-3 nearest training slides (cosine > threshold)." This issue implements exactly that mechanism.
- Repo evidence for the reuse pattern: `web_ui/src/lib/rag/rag-orchestrator.ts` already implements a fused-ranking pipeline this issue's ranking step should mirror rather than reinvent — [rag-orchestrator.ts:27](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/web_ui/src/lib/rag/rag-orchestrator.ts#L27) imports `rrfFuse` from `../search/rrf-fusion`, and [rag-orchestrator.ts:441-446](https://github.com/ZaxbyHub/trainingapp/blob/b2ddda06e86b63f81bdbedce220a3a2357bc9852/web_ui/src/lib/rag/rag-orchestrator.ts#L441-L446) shows the existing relevance-floor pattern (`const floor = didRerank ? MIN_CROSS_SCORE : MIN_RRF_SCORE; contextChunks = contextChunks.filter((c) => c.score >= floor);`) — this issue's cosine threshold for slide links should follow the same floor-then-filter convention for consistency, not introduce a third scoring convention.
- CONFIRMED training corpus scale for cost/complexity sizing: 384 slide documents (#77/D1) is the fixed candidate set every doc chunk in every active pack must be compared against — a bounded, cheap top-K search, not a full corpus-wide join.

- #79 (D3) supplies the built training pack (384 slide docs, prebuilt embeddings) that this issue's link computation targets. #70 (C3, Node PackManager) supplies the pack install/supersede lifecycle events this issue hooks the recompute trigger into. #65 (B7, hybrid retrieval) supplies the RRF/reranker infrastructure (`rrfFuse`, floor conventions) this issue's ranking step reuses.
- This issue OWNS the `links` table schema, the top-3 nearest-slide computation, and the recompute trigger wiring. It does NOT own pack install/supersede itself (#70) or the retrieval pipeline's own RRF fusion (#65) — only the training-specific cross-link derived from them.
- Not parallel-safe with #70/#65 (hard schema/pipeline dependency); parallel-safe with #81 (D5) once both depend on #79.

## Required scope
- [ ] `contracts/store.schema.sql` addition (coordinate with #59/B5, which owns this file): `links` table — proposed columns `chunk_id TEXT, slide_id TEXT, pack_id TEXT, score REAL, rank INTEGER, computed_at TEXT, PRIMARY KEY (chunk_id, slide_id)`.
- [ ] `packtool/links/compute-links.ts`: for each doc chunk, embed (or reuse existing chunk embedding) and compute cosine similarity against all 384 training slide embeddings; keep top-3 above a threshold (propose `links.cosineThreshold: 0.5` as a config default, consistent with the audit's cited "unrelated pairs <0.35" reference point — cite `audit_report.md` Sources for any such number and mark as to-be-measured against the eval set, #54/A4).
- [ ] Build-time computation: shipped doc packs (bundled, per C1/#68) get links computed during `packtool build-docs`/`verify` (coordinate with #69/C2, #73/C6) so installation ships pre-linked.
- [ ] Client-side recompute trigger: hook into #70's (C3) pack install/supersede lifecycle events to recompute links for newly-active or newly-superseded packs, without recomputing the entire table on every install (only affected chunk/slide pairs).
- [ ] Eval metric wiring: `eval/questions.jsonl` (#54/A4, once it exists) questions carrying an expected training slide id are the ground truth this issue's link quality is measured against — name #54 as supplying that fixture in prose, do not add it as a blocking dependency edge.
- [ ] Out of scope: Python-side link computation for the Python-backend path unless #69 (C2) explicitly requires parity in this cycle — if Python parity is needed, split into a follow-up rather than silently expanding this issue's scope.

## Acceptance
| Case | Required proof |
| --- | --- |
| Links computed at build time | Building a doc pack against the training pack from #79 populates non-empty `links` rows for chunks with a genuine topical match (fixture: a doc chunk whose text closely mirrors a known slide's on-screen text) |
| Recompute on pack install | Integration test: install a new doc pack after the training pack is active; assert `links` rows appear for the new pack's chunks without a manual trigger |
| Recompute on pack supersede | Superseding a pack removes/updates its stale `links` rows rather than leaving orphaned entries pointing at inactive chunk ids |
| Top-3 cap enforced | Unit test asserts no chunk has more than 3 `links` rows even when more than 3 slides exceed the threshold |
| Threshold respected | Unit test with a synthetic embedding pair below threshold asserts no link row is created |

## Evidence and exit gate
Closing PR must attach: the `links` table migration, the build-time and recompute-trigger test results, and (once #54/A4's eval set exists) a preliminary Learn hit@3 measurement using this table — final target validation belongs to #82 (D6), but this issue should report its own contribution number. Invalidated if #59's `store.schema.sql` changes shape after this table is added (re-migrate) or if #65's RRF/floor convention changes.

Treat PR 4 of 7 as a stable tracking slot, not merge order. The explicit dependencies govern execution; focused implementation issues supply evidence to later qualification issues without depending on those later issues closing. Re-anchor source and active-PR overlap before implementation. Close only the fully satisfied scope: wired production callers, meaningful focused validation and documented outcomes, with no silent deferral. Open one PR per issue titled with this issue's prefix; keep `contracts/` (API OpenAPI, store schema, pack schema) authoritative; add tests for every acceptance row; update docs touched by the change. Verification-only residuals may close with concrete current evidence rather than an unnecessary code patch.

| Slot | Issue | Work |
|---|---|---|

| Slot | Issue | Work |
|---|---|---|

| Slot | Issue | Work |
|---|---|---|

| Slot | Issue | Work |
|---|---|---|

| Slot | Issue | Work |
|---|---|---|

zaxbysauce:
Closes #80

## Summary

Implements D4 of the v3 roadmap (Workstream D): every document chunk from
every active pack gets its top-3 nearest training slides recorded in the
store's `links` table — computed at pack-build time for doc packs
(`packtool links`) and recomputed transactionally in-store whenever doc
content changes — so a citation to a vendor PDF can also point at the exact
training slide covering the same content. Finalizes the reserved `links`
table to schema v2 (pack_id/rank/computed_at + both migrate ladders), adds
links-integrity validation to `packtool verify`, and exposes the runtime
maintenance surface (`recomputeLinksForDocs` / `pruneOrphanLinks` /
`removeLinksForPack`) that the future pack lifecycle (#70) and build-docs
(#73) compose. Learn panel UI (#82) and the player (#81) are untouched, as is
Python-side link computation (per the issue's own scope line).

## Root Cause

The `links` storage demanded by D4 was reserved-but-dead schema: `contracts/store.schema.sql:84-91` created a minimal `links` table (chunk_id, slide_id, score — no pack_id/rank/computed_at) that no code wrote or read (repo-wide grep: zero statements). No build-time compute path existed (`packtool/cli.ts` registered only storyline/build-storyline/verify), no doc-pack builder exists yet (#73 open), no pack install/supersede lifecycle exists yet (#70 open), and the runtime ingest writer (`desktop/main/backend/ingest/pipeline.ts` `writeDocument`, lines 331-409) neither populated nor cleaned `links` — its DELETE-BEFORE-REINGEST cascade (lines 350-363) cleaned embeddings and chunks_fts for stale chunks but would have orphaned links. The reserved DDL was pinned at `meta.schema_version = '1'`.

## Fix

Own the links surface end to end at the seams that exist today, without duplicating #70/#73 scope:

- `contracts/store.schema.sql` — links finalized to the issue's shape: `chunk_id, slide_id, pack_id, score, rank, computed_at, PRIMARY KEY (chunk_id, slide_id)`; `meta.schema_version` seed bumped to `'2'` per the in-file SCHEMA-BUMP-CONTRACT.
- `desktop/main/backend/store/migrate.ts` + `contracts/tests/store-interop/migrate.py` — 1->2 ladder step (drop + recreate the v1 reserved table, lossless: it was write-free by construction); both runtimes stay interop-compatible (bidirectional suite passes).
- `packtool/links/compute-links.ts` (new, pure) — `computeDocSlideLinks`: cosine top-K (default 3) above a threshold (default 0.5, strict `>`, floor-then-filter convention), deterministic slide_id tie-break, ISO-8601 `computed_at`; plus `slideIdFromDocPath`.
- `packtool/links/link-pack.ts` (new) + `packtool/cli.ts` — `packtool links --pack <docPack> --training <trainingPack> [--threshold] [--top]`: computes and writes links into a doc pack's prebuilt index (zip or dir, atomic rewrite, entry dates preserved). Guards: refuses `source_class: "training"` doc packs and refuses non-comparable embedding spaces (model_id, dims, and `normalize: true` must all match). This is the build-machine step #73's build-docs will call as a library.
- `packtool/build/index-writer.ts` — `writePackIndex` accepts optional links rows; `installPackRows` copies them (install stays a lossless wholesale move).
- `packtool/build/verify.ts` — index stamp now derives from the shared `STORE_SCHEMA_VERSION` constant (was a hardcoded `'1'`), plus links-integrity validation: orphan chunk/pack references, top-3 cap, contiguous ranks with descending scores, cosine range, parseable `computed_at`.
- `desktop/main/backend/store/links.ts` (new) — runtime maintenance: `recomputeLinksForDocs` (delete-then-insert, idempotent, no-op when no training slides), `pruneOrphanLinks`, `removeLinksForPack`, `loadSlideVectors`; kernel mirrors the packtool module with a duplicated golden-vector parity test.
- `desktop/main/backend/ingest/pipeline.ts` — the writeDocument transaction now deletes stale links in the existing cascade and calls `recomputeLinksForDocs(db, [docId])` before COMMIT, so links are always transactionally consistent with chunks on the only doc-content mutation path that exists today; #70's future PackManager composes the same store-layer operations on install/supersede.
- Version pins updated deliberately: `b5-store` / `b6-*` suite expectations and `build-storyline.ac8` now expect schema v2 (assertions still pin seeds + version).
- `packtool/README.md` — documents the `links` verb.

Python-side link computation remains out of scope per the issue (except the migrate ladder the SCHEMA-BUMP-CONTRACT requires). Eval wiring is prose-only: `eval/questions.jsonl` `expected_training_slide_id` (currently null in all rows) is #54's fixture for a future Learn hit@3 measurement — this issue's contribution is the `links` table itself.

- Defect class: derived-table rows outliving the rows they reference (chunk_id-keyed tables with unenforced FKs must be cleaned by every writer that deletes chunks).
- Sweep result: 4 predicates, 28 hits, all dispositioned (10 FIX — the new/extended links maintenance in this diff; 18 FALSE_POSITIVE/OUT_OF_CLASS — pre-existing correct sites or base-table writers) — see the trace's 08a artifact.
- Guardrail: `verifyPack` now structurally validates shipped links rows (CI-enforced), and frozen check C3 proves the supersede cascade prunes orphans (RED on base: orphan row survived; GREEN post-fix). The guardrail bites: seven corruption fixtures (orphan chunk ref, ghost pack ref, cap violation, rank/score inversion, out-of-range cosine, bad timestamp) are all rejected in `links-verify-integrity.test.ts`.

## Test plan

- Regression test: `bash .agents/issue-traces/80-recompute-doc-slide-links/repro/replay-red.sh` (7 frozen acceptance checks via repro-check.sh) -> C1-C6 RED->GREEN PASS, C7 GREEN->GREEN PASS
- Impacted suite: `cd packtool && npx vitest run --config vitest.config.ts` -> 95/95 PASS
- Impacted suite: `cd desktop && npx vitest run --config vitest.config.ts` -> 358 passed / 3 pre-existing env-gated skips / 0 failed
- Cross-runtime: `python contracts/tests/store-interop/run_interop.py --direction "node->python"` and `"python->node"` -> both PASS (rows/topk/fts/hashes match) on the v2 schema
- Lint/type/build checks: `npx tsc -p tsconfig.json` (packtool) and `npx tsc --noEmit -p tsconfig.json` (desktop) -> PASS; packtool build gate exercised inside extract.ac8
- Deferred-work scan: `scan-deferred.sh` -> clean

## Regression Protection

- New: `packtool/storyline/__tests__/links-ac{1,4,5}-*.test.ts`, `links-preserve-roundtrip.test.ts` (frozen acceptance checks, authored by an independent context and frozen pre-implementation), plus `links-compute-unit.test.ts`, `links-cli.test.ts`, `links-verify-integrity.test.ts`.
- New: `desktop/src/__tests__/d4-links-ac{2,3}-*.test.ts`, `d4-links-schema.test.ts`, `d4-links-store.test.ts` (runtime trigger, supersede/orphan, schema v2, migrate ladder incl. fresh-vs-migrated shape parity pin).
- Boundary/negative coverage: threshold exactness (0.8 boundary), zero-norm and length-mismatch vectors, invalid options, tie-break determinism, 999+ slide pads, zip-vs-dir packs, model/dims/normalize mismatch refusals, training-pack-as-doc refusal, missing packs.
- Test drift review: schema-version pins in b5/b6/ac8 updated deliberately (documented in the trace's 08 artifact); no stale assertions left.

|---|---|
| AC1 Links computed at build time | Frozen check C1 (`links-ac1-buildtime.test.ts`) — real training pack via buildStorylinePack + doc-pack index via writePackIndex; mirror chunk links to S1 at cosine ~1, orthogonal chunk has no rows; PASS post-fix |
| AC2 Recompute on pack install | Frozen check C2 (`d4-links-ac2-install.test.ts`) — installPackRows(training) + IngestPipeline ingest with a matching vector; links rows appear without a manual trigger; base RED (0 rows) -> PASS |
| AC3 Recompute on supersede | Frozen check C3 (`d4-links-ac3-supersede.test.ts`) — re-ingest same path (DELETE-BEFORE-REINGEST); zero orphan links + new chunk linked; base RED (orphan survived) -> PASS |
| AC4 Top-3 cap enforced | Frozen check C4 (`links-ac4-topk.test.ts`) — 5 slides above threshold yield exactly 3 rows, ranks {1,2,3}, descending; PASS |
| AC5 Threshold respected | Frozen check C5 (`links-ac5-threshold.test.ts`) — cosine ~0.05 pair produces no row at the 0.5 default; PASS |
| Schema shape (required scope) | Frozen check C6 (`d4-links-schema.test.ts`) — links columns include pack_id/rank/computed_at, PK (chunk_id, slide_id), meta.schema_version = '2'; PASS |

## Invariant Audit

The repo has no dedicated invariant doc; audited against ARCHITECTURE.md and the ADR set:

- Per-profile shared SQLite store, schema frozen at `contracts/store.schema.sql`, cross-runtime interop proven (ARCHITECTURE.md:47-52, ADR-0005): touched — v2 bump performed through the file's own SCHEMA-BUMP-CONTRACT (seed + both migrate ladders); bidirectional interop suite passes on the v2 schema.
- Pinned sqlite-vec 0.1.9 (ADR-0005): not touched — no extension-version change.
- Profile-scoped store layout (ADR-0006): not touched.
- Relevance-floor calibration conventions (ADR-0007 + rag-orchestrator floor-then-filter): followed — links threshold 0.5 is a documented cosine floor on its own score scale with the strict floor-then-filter form; retrieval config untouched.

## Risk and Rollback

- Risk level: medium (authoritative cross-runtime schema change, mitigated by the mandated dual migrate ladders + interop suite + parity pins).
- Rollback: revert the single fix commit; v2 stores fail loud on older code (migrate's future-version guard) — acceptable for an unmerged feature branch; no shipped data migration to unwind.
- Residual risk: none known; pack install/supersede recomputation composes the shipped store-layer operations when #70 lands (the single call surface), and build-docs (#73) calls the same compute library.

## Waivers (or none)

none

## Merge status

PR head: a2840fb7bf3bbd071f94d4451f024cebc9f78926

## Issue Closure

Closes #80


zaxbysauce:
## Swarm PR Review — PR #109 (issue #80: D4 doc-to-slide links)

Structured review (6 base dimension lanes + 2 targeted risk-family micro-lanes + 3 independent reviewer passes + 1 critic challenge, all against `pr_head_sha` `a2840fb7`, `base_sha` `e09a6c6`). Full methodology: `swarm-pr-review` skill, Profile B (Claude Code native subagents), depth tier L.

### PR intent
Finalizes the reserved `links` table to schema v2 (`pack_id`/`rank`/`computed_at`), adds a `packtool links` CLI verb + library to compute top-3 nearest-training-slide links for doc packs at build time, and wires transactional runtime recompute into the ingest pipeline. Dual migrate ladders (Node + Python) extend the schema-bump-contract.

### Deterministic signals (independently re-run, not just trusted from the PR body)
| Claim | Result |
|---|---|
| packtool: 94/94 tests pass | ✅ Confirmed exactly |
| desktop: 356 passed / 3 skipped / **0 failed** | ⚠️ 356 passed / 3 skipped confirmed, but **1 failed** locally (`b2-csp-pin.test.ts`) |
| Cross-runtime interop both directions PASS | ✅ Confirmed (`rows_match`/`topk_match`/`fts_match`/`hashes_match`: true both ways) |
| `tsc` clean (packtool + desktop) | ✅ Confirmed, both exit 0 |
| CI status | ✅ All 10 required checks green; `mergeStateStatus: BLOCKED` is solely `REVIEW_REQUIRED` (branch protection wants an approval) — no CI or merge-conflict issue |
| Issue #80 closure integrity | ✅ Open, referenced only by this PR, no duplicate-closure risk |

The one test failure (`b2-csp-pin.test.ts`, a CRLF/LF eol assertion on `web_ui/index.html`) was traced to this review worktree's `core.autocrlf=true` converting a file **not in this PR's diff** to CRLF on checkout — the committed blob is LF, so this reproduces the author's/CI's "0 failed" claim faithfully in their environment. Not a PR regression; flagged only so the discrepancy isn't re-litigated.

### Confirmed findings

**F-TOPBUG (MEDIUM) — `packtool links --top <k>` is silently a no-op**
`packtool/cli.ts:262` builds the CLI options object as `{ top: args.top }`, but `LinkPackOptions`/`computeAndWriteLinks` (`packtool/links/link-pack.ts:45-48,179`) only read `options.topK`, defaulting to `LINKS_TOP_K=3` whenever `topK` is undefined. Independently found and empirically reproduced by three separate explorer lanes, confirmed by an independent reviewer, and re-verified by the critic — all four runs showed the CLI's own stderr echoing `top 3` while invoked with `--top 1`. `tsc` doesn't catch it (excess-property checks don't fire on a conditionally-spread key), and the existing regression test `links-cli.test.ts` `"honors --top 1"` cannot discriminate a working flag from the ignored default, because the fixture happens to yield exactly 1 qualifying slide either way.
- Blast radius is narrower than it first looks: `--threshold` is unaffected (maps correctly), the top-3 cap itself is proven at the kernel level with tests that call `computeDocSlideLinks` directly (so AC4/AC5 hold), the effective value is echoed to stderr on every run (not silent), and nothing in the codebase reads the `links` table yet.
- **Fix is not a safe one-liner.** `packtool/build/verify.ts:284` hard-codes a `≤3` cap and `contracts/store.schema.sql:101` documents `rank 1..3` as a cross-runtime contract. A bare `top` → `topK` rename would let `--top 5` build a pack that `packtool verify` then rejects. The fix needs to either reject `k > 3` at CLI parse time (alongside the existing `Number.isInteger`/`>=1` guard) or widen the verify cap and schema/README contract together — plus a test change that asserts the actual resolved `topK`, not just row count.

**F-CS02 (LOW) — `pruneOrphanLinks` doesn't cover the one orphan class that can occur**
`desktop/main/backend/store/links.ts:201-206` only deletes links whose `chunk_id` no longer exists — but `chunk_id` is FK-protected (`store.schema.sql:97`, FKs on by default; the module's own tests must explicitly `PRAGMA foreign_keys = OFF` to fabricate that case). `slide_id` (`:98`) has no FK and is the orphan class that can actually happen (e.g. a superseded training pack), and nothing prunes it. Currently unreachable — no production pack-install/supersede code exists yet (only test callers; deferred to #70) — so this is a documentation/contract gap, not a live bug: the module's "single runtime call surface for link maintenance" header comment should say slide-side orphans are explicitly out of scope for now. (An earlier pass also flagged `removeLinksForPack`'s `pack_id` semantics as mis-keyed; on review that's incorrect — `pack_id` is documented and correctly implemented as the *owning doc pack*, not the training pack, by design.)

**CS-01 (LOW)** — `recomputeLinksForDocs` (`links.ts:160-161`) has an early `continue` that skips the links `DELETE` for a doc with chunks-but-no-embeddings, contradicting its own "idempotent delete-then-insert" docstring. Unreachable via the only current production caller.

**CD-02 (LOW)** — `sqlite-store.ts:160` runs the v1→v2 migration unconditionally and irreversibly in place on every store open, no backup. Confirmed milder than first reported: rolling back to the pre-PR binary afterward hits a pre-existing (not part of this PR) silent-degrade path (`index.ts:619-624`) rather than data loss or a crash — the store just becomes unavailable with only a console log.

**UIB-002 (LOW)** — `link-pack.ts:255-260`'s bare `INSERT` (no `ON CONFLICT`) against `PRIMARY KEY (chunk_id, slide_id)` can abort with a raw SQLite error (correctly rolled back, not corrupting) if a malformed Storyline publish contains two slides sharing the same untrusted `slide.id`.

**MIGRATE-PY-DEAD (LOW)** — `contracts/tests/store-interop/migrate.py`'s `migrate()` (the Python v1→v2 ladder) has zero callers anywhere; `run_interop.py` applies the schema fresh at v2 rather than exercising the migration path. This PR adds a docstring claim ("the store-interop suite proves the two runtimes agree") that isn't backed by any executed test.

**INSTALLROWS-UNGUARDED (LOW)** — `packtool/build/index-writer.ts:239-241`'s new links `SELECT` has no try/catch (unlike `verify.ts`'s equivalent), so it would throw unhandled against a pre-PR v1 pack file. No production caller exists yet, and `verify.ts`'s schema-version check would reject a v1 pack before this code path is reached anyway — worth a `meta.schema_version` precondition check whenever #70 wires a real caller.

**IA-004 (informational)** — a comment in `compute-links.ts`/`links.ts` misattributes an inclusive (`>=`) rag-orchestrator precedent to this module's intentionally exclusive (`>`) threshold — comment-only, no functional effect.

### Disproved candidates (for completeness, not blocking)
- A prior-PR seeded hypothesis about unclosed `better-sqlite3` handles in `packtool/build/index-writer.ts` — **disproved**, the try/catch-close guard predates this PR.
- A concurrency-race hypothesis on `recomputeLinksForDocs`'s delete-then-insert — **disproved**, it runs inside `BEGIN IMMEDIATE`/`COMMIT`, giving transactional (not merely app-layer) protection.
- A skip-guard concern on the new D4 desktop test files — **disproved as PR-introduced**; identical to 14 pre-existing test files' convention, and all 14 new/affected tests execute (zero skips) with deps installed.

### Obligations
AC1–AC3, AC6 (schema shape): MET, each proven by its own frozen acceptance test. AC4 (top-3 cap) and AC5 (threshold): MET **at the kernel level** (direct `computeDocSlideLinks` tests), but AC4 is **not exercised correctly through the CLI** due to F-TOPBUG. The PR body's internal defect-class sweep ("4 predicates, 28 hits...") references a trace artifact not present in this repo/diff — UNVERIFIABLE, not disproved.

### Verdict: **REQUEST_CHANGES**

No CRITICAL or HIGH findings, and CI is fully green — this is a solid, well-tested implementation of D4. The one blocking item is **F-TOPBUG**: a real, PR-introduced, user-facing bug in a newly-documented CLI flag, confirmed independently four separate times. It's cheap to fix now and meaningfully more expensive to unwind later given its interaction with `verify.ts`'s hard-coded cap — exactly the kind of thing worth catching before merge rather than after. Everything else here (F-CS02, CS-01, CD-02, UIB-002, MIGRATE-PY-DEAD, INSTALLROWS-UNGUARDED) is LOW/informational and can reasonably ship as-is or as fast-follow issues alongside #70.

---

